PR #33, opened 2019-08-30
WIP Cards v2: Header image motion during route transition

burieberry:


burieberry:
Unresolved problem:
Inner box seems to move slightly faster than the outer box when transitioning from individual card page to list page (`/cards-v2/2` to `cards-v2`). See the yellow background of the outer box peeking through on the right side of the focused box:

card

The difference is especially obvious when using `scale` instead of `resize` during animation.

PR #43, opened 2019-09-18
Don't use magic number

bagby:


ef4:
This clarifies the problem a little better. I think it means we have a bug in ember-animated. 

This kind of behavior (compensating for the motion of your offsetParent) is supposed to always be built in.

PR #75, opened 2019-09-30
improved interruptions and simplified shadows

ef4:
@burieberry this cleans up most (though not all) the interruption behaviors in your branch, and it switches to the new boxShadow motion.

### DIFF
--- a/tests/dummy/app/controllers/form-cards/edit.js
+++ b/tests/dummy/app/controllers/form-cards/edit.js
@@ -4,8 +4,8 @@ import { action, set } from '@ember/object';
 // import { printSprites } from 'ember-animated';
 import move from 'ember-animated/motions/move';
 import resize from 'ember-animated/motions/resize';
-import opacity from 'ember-animated/motions/opacity';
 import { easeInAndOut } from 'ember-animated/easings/cosine';
+import boxShadow from 'ember-animated/motions/box-shadow';
 
 import { duration } from './index';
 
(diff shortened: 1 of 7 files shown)


tests/dummy/app/controllers/form-cards/edit.js
```diff
@@ -18,12 +18,18 @@ export default class FormCardsEditController extends Controller {
...
+    insertedSprites.forEach(sprite => {
       sprite.startAtPixel({ y: -1.5 * window.innerHeight });
       move(sprite, { easing: easeInAndOut, duration });
       sprite.applyStyles({ 'z-index': 2 });
     });
 
+    receivedSprites.forEach(sprite => {
+      move(sprite, { easing: easeInAndOut, duration });
+      sprite.applyStyles({ 'z-index': 2 });
+    });
```
ef4, line 30: insertedSprites and receivedSprites need different handling here because receivedSprites already have their own initial position. If we use `startAtPixel` on them they will jump during interruption.

tests/dummy/app/controllers/form-cards/index.js
```diff
@@ -1,36 +1,29 @@
...
   @action edit(card) {
-    for (let elt of this.model) {
-      if (elt.id === card.id) {
-        set(elt, 'expanded', !elt.expanded);
-        this.transitionToRoute('form-cards.edit', elt);
-      } else {
-        set(elt, 'expanded', false);
-      }
-    }
+    this.transitionToRoute('form-cards.edit', card.id);
```
ef4, line 17: The whole "expanded" property wasn't really doing anything here.

tests/dummy/app/controllers/form-cards/index.js
```diff
@@ -40,38 +33,28 @@ export default class FormCardsIndexController extends Controller {
...
     sentSprites.forEach(sprite => {
-      move(sprite, { easing: easeInAndOut, duration });
-      resize(sprite, { easing: easeInAndOut, duration });
-      opacity(sprite, { from: 0, easing: easeInAndOut, duration });
-      sprite.applyStyles({
-        'z-index': 3
-      });
+      opacity(sprite, { easing: easeInAndOut, duration });
+      scale(sprite, { easing: easeInAndOut, duration });
+      move(sprite, { disableMomentum: true, easing: easeInAndOut, duration });
```
ef4, line 56: The insertedSprites get an explicit `startAtPixel` and they get an explicit about of scaling via our `scaleBy`. 

But the `sentSprites` (which we will get during an interruption) keep their own initial position, and use `scale` because `scale` uses initialPosition and finalPosition, rather than jumping to any particular size.

PR #82, opened 2019-10-09
Transition from view to edit

### LINKED ISSUE #59: Card transitions: switching between view and edit modes
View mode is how the consumer (reader) sees the page. 
In the edit mode user (editor, admin, etc) can modify content and/or schema. 

When user hovers mouse over the card in view mode we display the border around card (same as in #54). Once user clicks `Edit`, the card is transmuted to Edit mode. Following transition happen in order: 

1. Card is transmuted. By default card is switching form View to Layout Edit (inline edit) mode. In the inline edit mode fields can be edited not in the right edge, but in place (editing in this view is subject of #61). The fields are displayed in the same way as in view mode, but card's aspect ratio, borders, and right edge and possibly other properties are changing. 
2. As card is resizing the selection border platter is hiding (see transition 1 in #55)
3. After card transformation to edit mode is compete, the edit mode border appears (see #56, #57) and right edge slides in from the right

When user switches from edit to view mode, the above transformations are reversed: 
1. Hide right edge, hide border
2. Transmute card to view mode

In View mode border will be displayed on hover (i.e. if user doesn't move mouse after clicking on "Done" we will not display the border immediately after transformation)

Each ui component is responsible for displaying its contents in relevant mode and transitions between modes. 

**Note:** If it turns out that the card is overcrowded with movement during transitions we might need to have an option specifying which elements we want to animate, and which to fade out/fade in

https://www.dropbox.com/sh/k7u4198p0sdrj4c/AAD7JvVVlZVE8LMALx-2nlq9a?dl=0&preview=04_Switch+View+to+Edit.mp4

robustdj:
Closes #59

### DIFF
--- a/tests/dummy/app/controllers/catalog/events/edit.js
+++ b/tests/dummy/app/controllers/catalog/events/edit.js
@@ -7,7 +7,6 @@ import adjustColor from 'ember-animated/motions/adjust-color';
 import { default as opacity } from 'ember-animated/motions/opacity';
 import { printSprites, wait } from 'ember-animated';
 import { easeInAndOut } from 'ember-animated/easings/cosine';
-import { animationDelay } from '../../catalog/events';
 
 export default class CatalogEventsEditController extends Controller {
   showModeMenu = false;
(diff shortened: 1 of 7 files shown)


alexspeller:
Closing stale PR

PR #96, opened 2019-10-16
Update to Ember 3.13 (Octane)

bagby:


addon/components/tray.js
```diff
@@ -1,10 +1,6 @@
+import Component from '@glimmer/component';
 import { action } from '@ember/object';
-import { tagName, layout as templateLayout } from '@ember-decorators/component';
-import Component from '@ember/component';
-import layout from '../templates/components/tray';
 
-@templateLayout(layout)
-@tagName('')
 export default class TrayComponent extends Component {
   expanded = false;
```
jenweber: There is no `this.set` in Glimmer components, and anything that gets mutated & should be rerendered needs `@tracked`. I think it should be:

```
import Component from '@glimmer/component';
import { action } from '@ember/object';
import { tracked } from '@glimmer/tracking';

export default class TrayComponent extends Component {
  @tracked expanded = false;
  trayAction() {}

  @action
  isolate() {
    this.expanded = true;
  }
}
```

See https://octane-guides-preview.emberjs.com/release/upgrading/editions/#toc_tracked-properties and https://ember-learn.github.io/ember-octane-vs-classic-cheat-sheet/

addon/components/boxel.js
```diff
@@ -23,35 +18,27 @@ export default class BoxelComponent extends Component {
...
   }
 
   get name() {
-    if (this._name) {
-      return this._name;
+    if (this.args.name) {
+      return this.args.name;
     }
 
     return `boxel-${this.elementId}`;
```
jenweber: This will always be undefined. Glimmer components do not have a wrapping element, therefore no default element ID. See https://octane-guides-preview.emberjs.com/release/upgrading/editions/#toc_outer-html

addon/components/boxel.js
```diff
@@ -23,35 +18,27 @@ export default class BoxelComponent extends Component {
...
   }
 
-  set name(name) {
-    this._name = name;
-  }
-
   clickAction() {}
 
   transition = function*({ sentSprites, receivedSprites }) {
```
jenweber: GitHub won't let me add a comment in the section you didn't change, but you can't use `this.set` in Glimmer components. See the comments on tray.js

jenweber (changes requested):
In a manual test clicking through the dummy app, things seem to work (except for bugs that are out of the scope of this PR). However there are a couple uses of `this.set` to replace for these components to be valid in Glimmer.

It is tough to see if the variable scope is correct without having more tests. I think we have a decision to make about whether this PR should include those, or that should be separate work.

app/templates/components/tray.hbs
```diff
@@ -0,0 +1,3 @@
+<div role="button" {{action this.trayAction}} class="boxel-tray {{if this.expanded "expanded"}}">
+  {{yield}}
+</div>
```
jenweber:
```suggestion
<button {{on "click" this.trayAction}} class="boxel-tray {{if this.expanded "expanded"}}">
  {{yield}}
</div>
```
There are two optional changes shown here.

One is, in Octane, the `on` helper is preferred to action. However action still works fine, I think.

Second, `div role="button" causes a11y problems. See #99. This is optional because this was not a change your PR introduced and so it could be considered out of scope.

addon/components/tray.js
jenweber: This may also be useful https://octane-guides-preview.emberjs.com/release/state-management/when-to-use-get-and-set/

jenweber (approved):
I clicked through the dummy app after the latest updates, and it all still seems good!

PR #101, opened 2020-04-13
Adds initial seed data, routes, components and styling

burieberry:
- update dependencies and fix linting errors
- move old stuff elsewhere
- adds seed data
- movie-registry, view, and edit routes 
- movie and field components


alexspeller (approved):
This all looks good to me I think!

PR #104, opened 2020-04-23
Add card select component

alexspeller:
Also do some dependency updates and add ember-dependency-lint to prevent getting the same ember addon multiple times


burieberry (approved):
This looks awesome!

PR #107, opened 2020-04-29
Add imported data from google sheets / google drive

alexspeller:


burieberry (approved):
We can compress images or add smaller files if it causes any issues

PR #114, opened 2020-05-15
Fix deploy assets

alexspeller:
Fix assets in deploy.

As part of this, I've configured the build to use the rootURL `/boxel` in development too, so that it's immediately obvious when assets break.

If this is a pain, let me know and I'll pull that part out.

Deploy is working at https://cardstack.github.io/boxel/#/media-registry and it will deploy on merges to master.


burieberry (approved):
I think that's fine. We can change later if needed.

PR #197, opened 2020-06-30
Add page title that updates when page changes

alexspeller:
image


burieberry:
@alexspeller if you run `ember dependency-lint` you can replicate the errors locally. It looks like we need to upgrade ui-components, or allow different versions for failing deps

PR #243, opened 2021-01-12
Improve docs output for namespaced components

lukemelia:
Before:

Screen Shot 2021-01-12 at 4 55 27 PM

After:

Screen Shot 2021-01-12 at 4 53 53 PM


mattmcmanus (approved):
Nice

PR #244, opened 2021-01-12
Boxel::ProgressIcon component

lukemelia:
Open question for @burieberry and @mattmcmanus -- how should we handle the naming of the light-theme CSS class? 
I used `boxel--light-theme` but I'm sure that is not the BEM way.

https://user-images.githubusercontent.com/353/104303061-8b495080-5504-11eb-887a-728ce8f23624.mov

Screen Shot 2021-01-12 at 6 08 18 PM

### DIFF
--- a/app/components/boxel/progress-icon/index.hbs
+++ b/app/components/boxel/progress-icon/index.hbs
@@ -1,0 +1,11 @@
+<div
+  class="progress-icon {{if @isCancelled "cancelled"}} {{if @isComplete "complete"}}"
+  style={{this.elementStyle}}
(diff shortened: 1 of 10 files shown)


burieberry:
@lukemelia Looks good to me. We can try it the way you have it here and revisit/clarify with Gardella if there're any conflicts later on as we come across more examples. 

Also something to note (it's in the comments at the end of variables.css file) , #00B2AE is a color we've only used here because the usual boxel-highlight cyan color was too bright when small on light background. We can add it to the boxel color palette if more use cases arise.

PR #247, opened 2021-01-13
Extract Button and SelectionControlGroup components

lukemelia:
- Introduce [ember-class-names-helper](https://github.com/PrecisionNutrition/ember-class-names-helper)
- Add css-vars helper (same implementation as in #249)
- Boxel::DropdownButton replaces MediaCollection::Button
- Boxel::Button accepts a boolean @collectionStyle arg to apply styles for buttons in the collection view
- Boxel::Button yields a hash making it easy to add a "..." more button with dropdown
- SelectBtn renamed to Boxel::SelectButton

https://user-images.githubusercontent.com/353/104460752-74c8f500-55e9-11eb-8f97-a2597700d16d.mov

https://user-images.githubusercontent.com/353/104460749-74305e80-55e9-11eb-9404-df66286f2171.mov

https://user-images.githubusercontent.com/353/104460743-72ff3180-55e9-11eb-94c6-8154ae5f8523.mov

Screen Shot 2021-01-18 at 3 07 32 PM

Screen Shot 2021-01-18 at 3 07 46 PM


mattmcmanus (approved):
This is great

burieberry:
@lukemelia For the last one, I would only suggest to keep the styling the same as it was before when using on the collections pages. The text needs to be left-aligned and the width is smaller. The "..." icon needs to be more vertically centered, as in the image below.

btns-2

It might seem nit-picky, but we want the styling to be as exact as possible to the original Gardella design. I try to include a `@class` arg on every component to allow for customization.

lukemelia:
@burieberry cool, I'll take another pass over it tomorrow

lukemelia:
>It might seem nit-picky, but we want the styling to be as exact as possible to the original Gardella design. I try to include a @class arg on every component to allow for customization.

Good eye on this. I appreciate the attention to detail. Regarding `@class=`, is there a reason to do it that way instead of `class=` with `...attributes`?

burieberry:
@lukemelia At some point, passing in `class=` directly was overriding the whole class field, but it doesn't seem to be that way anymore. So yes, let's do it that way, then, much cleaner.

lukemelia:
@burieberry I've taken another pass, which I believe is faithful to the previous implementation's layout and styles, except for some purposeful improvements around hover states. Regarding the `@class` question, there is still one remaining use case where it is needed: when a component is invoked via the component helper and you want to add a class to it.

burieberry (approved):

PR #248, opened 2021-01-13
Missing component namespace and deprecated helper fix

burieberry:
- Fix missing namespace on breadcrumbs component on collection pages (https://cardstack.github.io/boxel/#/scenarios/media-registry/bunny_records/collections/golden-girl is currently broken)
- Use `includes` instead of `contains` helper, since it seems to be deprecated in ember-composable-helpers


lukemelia (approved):
👍

PR #249, opened 2021-01-13
Refactor CoverArt component to use CSS Variables

mattmcmanus:


lukemelia (changes requested):
It's a definite improvement from a code perspective. Having the logic contained within the component instead of having to rely on helpers is excellent. I don't think the layout would look very good without the modifier running though (i.e. fastboot), so I think that is an issue. I wonder if we could do a first pass at the calculation in the constructor of the component. That would run in fastboot. Not sure if the css var setting would work though.

mattmcmanus:
@lukemelia The need to support fastboot switched this to something arguably better. I switched css-var from an element modifier to a helper that be haves similar to our classnames helper. It sets the variables as part of the style tag, rather than using JS. The is identical to what the JS was doing, we're just writing it as an attribute. 

If my understanding of fastboot is correct, then this should work.

package.json
```diff
@@ -62,6 +62,7 @@
     "ember-load-initializers": "^2.1.1",
     "ember-math-helpers": "^2.14.0",
     "ember-maybe-import-regenerator": "^0.1.6",
+    "ember-modifier": "^2.1.1",
```
lukemelia: We can drop this again for now then, right?

PR #251, opened 2021-01-18
Adopt Freestyle::Usage

mattmcmanus:
CS-53


lukemelia:
This appears to break the `Source` output. It may be that we have to bake this into freestyle in order to fix that.

Screen Shot 2021-01-19 at 5 31 56 PM

app/components/boxel/breadcrumbs/usage.hbs
```diff
@@ -1,28 +1,17 @@
...
-      <div class="breadcrumbs__item-label">Collection</div>
-      <div class="breadcrumbs__item-title">Master Recordings</div>
-    </LinkTo>
-    <span class="breadcrumbs__carot" />
-    <LinkTo @route="media-registry.collection.index" @models={{array "bunny_records" "golden_girl"}} class="breadcrumbs__item">
-      <div class="breadcrumbs__item-label">Collection</div>
-      <div class="breadcrumbs__item-title">Golden Girl</div>
-    </LinkTo>
-  </Boxel::Breadcrumbs>
-</FreestyleUsage>
```
lukemelia, line 28: I think it is useful to show a "manually" constructed version, in case someone needs to do it. Is the ComponentUsage API conducive to showing more than one example?

app/components/doc/component-usage/args/index.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="ComponentUsageArgs">
+  <h3 class="ComponentUsageArgs__title">Component API</h3>
+
+  {{yield (hash
+    String=(component 'doc/component-usage/string')
+    Number=(component 'doc/component-usage/number')
+    Bool=(component 'doc/component-usage/bool' )
+    Action=(component 'doc/component-usage/argument' type="Action")
+    Yield=(component 'doc/component-usage/argument' type="Yield")
```
lukemelia: Seems like we also need `Object`, `Date`, `Component`, `Array` in order to cover know use cases in this repo (including open PRs).

app/components/doc/component-usage/argument/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+
+  {{#if (and hasBlock (not @hideControls))}}
+    <div class={{cn
+      "ComponentUsageArgument__controls"
+      (concat "ComponentUsageArgument__controls--" @type)
+    }}>
+      {{yield}}
+    </div>
+  {{/if}}
+</div>
```
lukemelia: +1 to a table or grid.

app/components/doc/component-usage/argument/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+
+    {{#if (is-present @defaultValue)}}
+      <em>(Defaults to: {{@defaultValue}})</em>
+    {{/if}}
+  </div>
+
+  {{#if (and hasBlock (not @hideControls))}}
+    <div class={{cn
+      "ComponentUsageArgument__controls"
+      (concat "ComponentUsageArgument__controls--" @type)
```
lukemelia: It is cool that Argument accepts a block so that you could provide a custom UI for a particular situation.

app/components/doc/component-usage/string/index.js
```diff
@@ -0,0 +1,13 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class ComponentUsageStringComponent extends Component {
+  @action onTextareaInput(event) {
+    let { value } = event.target
+    if (value) {
+      value = [value.split(',').filter(v => v)].flat();
+
+    }
```
lukemelia: Fine for now, but I'm not crazy about this UI solution.

app/styles/components/component-usage.css
```diff
@@ -0,0 +1,63 @@
+/*
+  ComponentUsage
+*/
+.ComponentUsage {
```
lukemelia: BEM naming would be `component-usage` IIRC

lukemelia (changes requested):

lukemelia:
a row hover state on the argument entries would be a nice touch, to make it easier to eye-track from the component to the UI widget.

app/components/boxel/breadcrumbs/usage.hbs
mattmcmanus: Right now we can just call ComponentUsage again

app/components/doc/component-usage/string/index.js
mattmcmanus: Added details to a subtask on the Linear issue: https://linear.app/cardstack/issue/CS-53/improve-component-documentation-by-declaring-arguments

lukemelia:
boxel

burieberry (approved):
This is great

PR #257, opened 2021-01-25
Thread banners

burieberry:
Added milestone banners with reveal animation. Also some basic styling for the dashboard in the background.

banners

### DIFF
--- a/app/components/boxel/thread-message/index.js
+++ b/app/components/boxel/thread-message/index.js
@@ -13,22 +13,14 @@ export default class ThreadMessageComponent extends Component {
   *transition({ insertedSprites }) {
     // printSprites(arguments[0]);
 
-    for (let [i, sprite] of [...insertedSprites].entries()) {
-      if (i === 0) {
-        yield wait(1200);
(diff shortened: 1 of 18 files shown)


lukemelia:
Looking good so far. Is this ready for review or WIP, @burieberry?

app/components/boxel/thread-message/index.js
```diff
@@ -15,7 +15,12 @@ export default class ThreadMessageComponent extends Component {
 
     for (let [i, sprite] of [...insertedSprites].entries()) {
       if (i === 0) {
-        fadeIn(sprite, { easing: easeOut, duration: 200 });
+        yield wait(1200);
```
lukemelia: Is this a "demo" feature? If so, should we find a way to pass in a flag to include this delay rather than hardcode it?

lukemelia (approved):

app/components/boxel/thread-message/index.js
burieberry: Managed to remove this extra wait time. Otherwise, there's a way to pass in a `duration` arg.

PR #258, opened 2021-01-26
Component CSS Co-location

lukemelia:
Adds a broccoli transformation to add an import to component JS files that have a colocated matching css file CS-13

- Synthesizes template-only component JS so that it works with both JS-backed components AND template-only components
- Looks for files named the same with a ".css" extension. e.g. `app/components/boxel/breadcrumbs/index.js/hbs` would import `app/components/boxel/breadcrumbs/index.css`. This is a change from previous pattern of naming the file "style.css" but makes this feature follow the colocated components conventions more.


lukemelia:
@ef4 Is importing `SynthesizeTemplateOnlyComponents` from embroider OK, or should I copy and paste that code since it is private to embroider?

ef4:
I think it's fine. It's true that it's private API, but (1) it's easy enough to copy paste later if we break it, (2) you're just as likely to accidentally pick up a bugfix from embroider than pick up breakage.

burieberry (approved):
Nice 👍

mattmcmanus (approved):
Love it

PR #260, opened 2021-01-26
Extract Boxel::ParticipantsSummary component

lukemelia:
CS-95

Screenshot 2021-01-27 at 12 54 04 AM
Screenshot 2021-01-27 at 12 52 44 AM


burieberry (changes requested):
I noticed the page breaks on the route that resolves to `/scenarios/workflow/hlc/1` since there's something undefined. Otherwise it looks good.

PR #265, opened 2021-02-02
Extract and document Boxel::IconButton

lukemelia:
Screenshot 2021-02-03 at 1 13 13 AM

Screenshot 2021-02-03 at 1 14 13 AM

### DIFF
--- a/app/components/boxel/icon-button/usage.hbs
+++ b/app/components/boxel/icon-button/usage.hbs
@@ -1,0 +1,44 @@
+<Freestyle::Usage
+  @name="IconButton"
+>
+  <:example>
+    <Boxel::IconButton
+      @icon={{this.icon}}
+      @variant={{this.variant}}
+      @width={{this.width}}
+      @height={{this.height}}
+      aria-label="Special Button"
+      {{on 'click' (fn this.log 'Button clicked')}}
+    />
+  </:example>
+
+  <:api as |Args|>
+    <Args.String @name="icon"
+      @description='the name of the svg to show'
+      @value={{this.icon}}
+      @onInput={{fn (mut this.icon)}}
+    />
+    <Args.String @name="variant"
+      @description='the variant to render as (applies CSS class) - `null` or `"primary"` or `"secondary"`';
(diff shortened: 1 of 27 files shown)


app/css/components/next-steps.css
```diff
@@ -1,8 +1,8 @@
 .next-steps {
   padding: 30px;
   background-color: #41404D;
-  color: var(--boxel-light);
-  font: var(--boxel-font-sm);
+  color: var(--boxel-utloight);
+  font: var(--boxel-utfoont-sm);
```
burieberry: Seeing some find + replace victims here

burieberry:
Also noticed that the base styling is overriding the use case styling here. Might have to do something like `.discrepancy-field__btn-container > .discrepancy-field__btn` in the discrepancy css file.

css-order

app/css/components/next-steps.css
lukemelia: Fixed, or should i say "utfoixed"

lukemelia:
> Also noticed that the base styling is overriding the use case styling here. Might have to do something like .discrepancy-field__btn-container > .discrepancy-field__btn in the discrepancy css file.

After considering a few options, I introduced a new component to avoid this.

app/components/boxel/icon-button/usage.hbs
```diff
@@ -0,0 +1,44 @@
...
+  </:example>
+
+  <:api as |Args|>
+    <Args.String @name="icon"
+      @description='the name of the svg to show'
+      @value={{this.icon}}
+      @onInput={{fn (mut this.icon)}}
+    />
+    <Args.String @name="variant"
+      @description='the variant to render as (applies CSS class) - `null` or `"primary"` or `"secondary"`';
```
mattmcmanus, line 22: Seems like passing in @options would be preferable here?

lukemelia: @mattmcmanus I wasn't sure how we would deal with null in that case, which is the most common and default value

burieberry (approved):

PR #268, opened 2021-02-03
Boxel::Menu component

lukemelia:
- menu-item helper for expressive menu definition
- add ember-link addon dependency to make creating link menu items easy
- Extract Boxel::SortMenu component

Boxel::SortMenu:

Screenshot 2021-02-02 at 7 01 06 PM

https://user-images.githubusercontent.com/353/106591458-422b7000-6589-11eb-9e22-abc46d3bc4ec.mov

Boxel::Menu:

https://user-images.githubusercontent.com/353/106883283-c2350f80-671a-11eb-8e66-45de22d21f1f.mov

Boxel::DropdownButton:

Screenshot 2021-02-04 at 6 55 00 PM

CS-142


burieberry:
@lukemelia Looks good but the scroll seems to be broken. Comparison with the current version:

https://user-images.githubusercontent.com/16160806/106924400-073a5100-66dd-11eb-94a7-d307c5d7750b.mp4

lukemelia:
> @lukemelia Looks good but the scroll seems to be broken. Comparison with the current version:

I'll take a look tomorrow

lukemelia:
I've addressed the SortMenu scrolling and made it better than the original. @mattmcmanus can you give this a review?

app/components/boxel/sort-menu/index.js
```diff
@@ -0,0 +1,15 @@
+import Component from '@glimmer/component';
+import { htmlSafe } from '@ember/template';
+export default class SortMenuComponent extends Component {
+  get styleAttribute() {
+    return htmlSafe(`max-height: ${this.maxHeight}`);
+  }
+
+  get maxHeight() {
+    let maxHeightArg = this.args.maxHeight;
+    if (maxHeightArg && maxHeightArg > 150) {
```
mattmcmanus: Can we make these consts with some clear names as to what they are?

mattmcmanus (changes requested):
Looks good to me. Just one request for clarification in the code.

mattmcmanus:
Also, are we OK with building on top of basic dropdown and wormhole? Should we consider something that may be a more better for the future? ie: popper.js?

mattmcmanus:
And it's higher level tool tippy.js: https://atomiks.github.io/tippyjs/

app/components/boxel/sort-menu/index.js
lukemelia: ✅

lukemelia:
> Also, are we OK with building on top of basic dropdown and wormhole? Should we consider something that may be a more better for the future? ie: popper.js?

I think that ember-basic-dropdown is a reliable primitive for the Ember community. When it is time to shift from wormhole, Miguel Camba will do it and have a migration path if necessary.

burieberry (approved):

PR #271, opened 2021-02-08
Extract Boxel::ActionChin

lukemelia:
CS-105

https://user-images.githubusercontent.com/353/107333477-802c1500-6af0-11eb-8ed4-cd3a40d8eae7.mov


burieberry (approved):
This is good for now. There'll be cases where the chin has multiple action buttons. We can discuss how to handle that when it comes up in later iterations.

PR #272, opened 2021-02-09
Refactor progress circle

Aierie:
CS-178

**Why?**
Previous component was rendering HTML based on variables within component, and geometry from an SVG file 'imported' by `svg-jar`. This produced two sources of truth that could conflict if geometry/variables changed on either side

**How?**
Extract two important variables that determine the circle's geometry - `outerCircleRadius`, and `progressArcThickness`, then define other variables based on those. Programmatically generate the svg, rather than 'importing' it, so that both svg and surrounding html are consistent with each other.

**Diagram of variables**
Screenshot from 2021-02-09 13-24-06

**Before**
https://user-images.githubusercontent.com/39579264/107321430-764ce680-6add-11eb-8e1d-93b40f6aad4c.mp4

**After**
https://user-images.githubusercontent.com/39579264/107321442-7cdb5e00-6add-11eb-949f-49a247fa2172.mp4

### DIFF
--- a/app/components/boxel/progress-circle/index.css
+++ b/app/components/boxel/progress-circle/index.css
@@ -1,5 +1,4 @@
 .progress-circle {
-  --icon-color: var(--boxel-highlight);
   display: flex;
   align-items: center;
   justify-content: center;
@@ -13,6 +12,16 @@
   width: 100%;
   height: 100%;
   transform: rotate(-90deg);
+}
+
+.progress-circle__background-circle { 
+  stroke: var(--boxel-light-400);
+  fill: none;
+}
+
+.progress-circle__indicator-circle {
+  stroke: var(--boxel-highlight);
+  fill: none;
   stroke-dasharray: 0 360;
   transition: stroke-dasharray var(--boxel-transition);
 }
--- a/app/components/boxel/progress-circle/index.css
+++ b/app/components/boxel/progress-circle/index.css
@@ -5,9 +5,6 @@
   justify-content: center;
   position: relative;
   margin: 0 auto;
-  background-position: center;
-  background-repeat: no-repeat;
-  background-size: contain;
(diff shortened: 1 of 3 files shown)


app/components/boxel/progress-circle/index.css
```diff
@@ -1,14 +1,9 @@
 .progress-circle {
-  --icon-color: var(--boxel-highlight);
```
lukemelia, line 2: How about keeping this, renaming it to `--progress-arc-color`, and using it in `.progress-circle__indicator-circle`? That way, a consumer could change the color of the progress arc by setting `--progress-arc-color` on the component's outer element.

app/components/boxel/progress-circle/index.css
```diff
@@ -17,6 +12,16 @@
...
+
+.progress-circle__background-circle { 
+  stroke: var(--boxel-light-400);
+  fill: none;
+}
+
+.progress-circle__indicator-circle {
+  stroke: var(--boxel-highlight);
+  fill: none;
   stroke-dasharray: 0 360;
```
lukemelia:
```suggestion
  stroke-dasharray: 0 360; /* Note: this is overridden by an inline style to show a progress arc of the right length */
```

lukemelia (approved):
I left 2 suggestions. And I love the explanatory diagram.

Aierie:
Thanks @lukemelia , I've made those changes. Also exposed a `--progress-arc-background-color` so that people can change the background arc color to match their custom `--progress-arc-color`.

mattmcmanus (approved):
This is a thing of beauty

burieberry (approved):
Awesome. I'm also happy that none of the styles are broken while making these great improvements. 👍  Go ahead and squash + merge

PR #273, opened 2021-02-09
Extract Boxel::Sidebar and Boxel::Header

lukemelia:
- Refactor ParticipantsBox and ProgressBox usage to use Boxel::Sidebar

Screenshot 2021-02-11 at 5 11 06 PM
Screenshot 2021-02-11 at 5 10 25 PM
Screenshot 2021-02-11 at 5 10 07 PM
Screenshot 2021-02-11 at 5 09 46 PM
Screenshot 2021-02-11 at 5 09 35 PM


app/components/boxel/thread-modal/index.hbs
```diff
@@ -30,24 +30,38 @@
...
-            @participants={{@model.participants}}
-          />
-          <Boxel::ParticipantList
-            @participants={{@model.orgRepresentatives}}
-          />
-        </:list>
-      </Boxel::ParticipantsBox>
+    <:sidebar as |SidebarSection|>
+      <SidebarSection
+        @header={{if @model.workflow.title (html-safe (concat "Workflow:<br>" @model.workflow.title)) 'Workflow Progress'}}
```
burieberry: The single and double quotes are mixed up here. It's " (double) for hbs and ' (single) for js.

app/css/templates/media-registry/item.css
```diff
@@ -1,31 +1,15 @@
-.item__sidebar {
-  padding-right: 30px;
+.item_sidebar {
+  margin-right: 30px;
```
burieberry: I think the adjustment for a margin that's outside of a component belongs to the outer component that holds it. If this is not a real padding that belongs inside of this component,  then it's better to make it a padding that belongs to that outer component. Since margins cause margin-collapse, it's safer to use padding when it's not certain that the margin won't collapse.

app/css/variables.css
```diff
@@ -22,7 +22,21 @@
...
+  --boxel-s-4: calc(var(--boxel-s-3) / var(--boxel-ratio));
+  --boxel-s-3: calc(var(--boxel-s-2) / var(--boxel-ratio));
+  --boxel-s-2: calc(var(--boxel-s-1) / var(--boxel-ratio));
+  --boxel-s-1: calc(var(--boxel-s0) / var(--boxel-ratio));
+  --boxel-s0: 1rem;
+  --boxel-s1: calc(var(--boxel-s0) * var(--boxel-ratio));
+  --boxel-s2: calc(var(--boxel-s1) * var(--boxel-ratio));
+  --boxel-s3: calc(var(--boxel-s2) * var(--boxel-ratio));
+  --boxel-s4: calc(var(--boxel-s3) * var(--boxel-ratio));
+  --boxel-s5: calc(var(--boxel-s4) * var(--boxel-ratio));
```
burieberry: This is a good idea and has been on my watch list to see which spacing amounts we use the most often. I have a general idea by now but need to double check and adjust calculation as needed. Also thinking of a clearer naming.

tests/integration/helpers/percent-complete-test.js
```diff
@@ -0,0 +1,35 @@
+import { module, test } from 'qunit';
```
burieberry, line 1: yay tests!

burieberry (commented):
So the componentization makes sense, but the styling is broken as you can see in side-by-side comparisons below.

The styling with the header with background color is normally reserved for embedded cards, and they have different styling attributes. Otherwise we can use the sidebar and section components. In the screenshot from the XD file below, there's another use case where the sidebar component is also used for something like the "Need Help?" container. We need to keep the base styling to a minimum and allow for variations without breaking the styling in other places. 

This is a good example for why we need the scoping feature of css-blocks more and more. I'll try it out and see if it resolves our problems.

sidebar

sidebars-2

This doesn't look right:

sidebar-usage

lukemelia:
@burieberry Great feedback! I'm embarrassed to realize how bad the color calibration is on my external display. It's so washed out that I couldn't even see the header background! I've improved the calibration and will also check my visuals on my macbook screen going forward. And will get this PR fixed up shortly.

app/css/variables.css
lukemelia: It would be great to get this set soon, and a font size scale as well. In addition, we should get Chris Gardella's buy-in on it, too, so that we can stick to the scale for the vast majority of spacing and font sizes. I took these names from what EveryLayout uses ("s" for "spacing", and then -4 to 4), but I'm definitely open to other ideas.

app/css/templates/media-registry/item.css
lukemelia: Will do. 👍

app/components/boxel/thread-modal/index.hbs
lukemelia: 👍

lukemelia:
@burieberry Updated.

PR #274, opened 2021-02-11
ActionContainer layout updates

burieberry:
Fixes [CS-187](https://linear.app/cardstack/issue/CS-187/general-updates-to-action-container-design-also-applies-to-layout)

I did another pass at this using modular spacing, and also updated the prepaid card styling.

Changes:
- Fix padding for prompt and content
- Layout changes
- Header styling
- Focus and hover styles
- Use modular spacing variables
- Prepaid card styling

action-1
action-2
layout-1
layout-2


lukemelia (approved):
Looks good. I like the shift to `rem` units in general, but would prefer named vars like `--boxel-spacing` rather than decimal `rem` values everywhere.

PR #276, opened 2021-02-16
CS-182 adjustments for sidebar

burieberry:
sidebar-with-related-content

@lukemelia I started fixing the sidebar styles, and I've only got the media-registry so far. (The CoverArt component needed some changes, too.) I'd need to spend at least another day refactoring the sidebar for the thread components and try to make it work for both Card Pay and the other workflow container types. I've already tested this in the past, and I wasn't happy with the results. So I recommend *not* using this component for the thread container.


app/css/variables.css
```diff
@@ -22,20 +22,18 @@
...
-  --boxel-s-3: calc(var(--boxel-s-2) / var(--boxel-ratio));
-  --boxel-s-2: calc(var(--boxel-s-1) / var(--boxel-ratio)); /* 8.43px */
-  --boxel-s-1: calc(var(--boxel-s0) / var(--boxel-ratio)); /* 11.6px */
-  --boxel-s0: 1rem;
-  --boxel-s1: calc(var(--boxel-s0) * var(--boxel-ratio)); /* 22px */
-  --boxel-s2: calc(var(--boxel-s1) * var(--boxel-ratio)); /* 30.25px */
-  --boxel-s3: calc(var(--boxel-s2) * var(--boxel-ratio)); /* 41.6px */
-  --boxel-s4: calc(var(--boxel-s3) * var(--boxel-ratio)); /* 57.2px */
-  --boxel-s5: calc(var(--boxel-s4) * var(--boxel-ratio));
+  --boxel-ratio: 1.375; /* calc(22/16) */
```
lukemelia: This ratio is close to the perfect fourth (1.333) and the augmented fourth (1.414). Once implemented, it would be cool to try those out and see if they give things a different feel.

app/components/boxel/cover-art/usage.hbs
```diff
@@ -27,6 +29,16 @@
...
+    <Args.Number
+      @name="displayCount"
+      @defaultValue="all covers in the given array"
+      @min={{1}}
+      @max={{10}}
+      @step={{1}}
+      @description="the number of covers displayed"
+      @value={{this.displayCount}}
+      @onInput={{fn (mut this.displayCount)}}
+    />
```
lukemelia: An alternative to adding this arg would be to use [`take` from ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers#take): `@covers={{take 5 this.coverThumbs}}`. I think it's slightly better to use composition for this, but don't feel strongly about it.

app/components/boxel/header/index.css
```diff
@@ -1,17 +1,23 @@
...
-  --header-padding: var(--boxel-s-1);
-  --header-background-color: var(--boxel-purple-100);
-  --header-min-height: var(--boxel-s2);
-
-  background-color: var(--header-background-color);
+  /* customizable css variables:
+    --header-padding
+    --header-min-height
+    --header-background-color
+  */
```
lukemelia: Why is it better to use the fallback style, rather than setting a default value? (I'm guessing it has something to do with the cascade, but am curious.)

lukemelia (approved):
Great improvements. I like the spacing var names.

app/components/boxel/cover-art/usage.hbs
burieberry: Didn't realize there was a helper for that, very helpful, thanks!

app/components/boxel/header/index.css
burieberry: @lukemelia Yes, probably because of the order these stylesheets are processed, the variable in this file was overriding it when I used it elsewhere. It would be ideal to not have to do that.

burieberry:
I'll make the changes mentioned above, as well as fixing the previous variable names used in other places so that they also work for the other sidebar components.

burieberry:
@lukemelia This is ready for another review. (Also, rebased this branch with master because it was red, but it might have done things worse... Maybe if you rebase your branch, we can merge this.)

New changes:
- I let the media-registry embedded-card styles alone, since those represent cards and are used in other places around the music registry app.
- Created Sidebar::CardContainer component for those white sidebar boxes. They reside within the SidebarSection component. (I named it that way because we named ActionContainer similarly, but feel free to change it.)
- Sidebar boxes aren't really using the same header component as the card headers, it's just the same font styling, so I used the Header component only for the cards.

Here's the updated documentation with the three sidebars:

sidebars

source

Use cases:

e-commerce-thread

card-pay-thread

Sidebar, SidebarSection, Header components:
media-reg

PR #281, opened 2021-02-18
Correct CSS url()s so that they work in dev and production

mattmcmanus:
Since we prepend `/boxel/` in production, all asset urls need to be relative. 

Please read the extended descriptions on the commits to make sense of why the relative urls are as 🤪 as they are


tests/dummy/app/css/templates/media-registry/master-collection.css
```diff
@@ -16,7 +16,7 @@
   right: 5px;
   width: 30px;
   height: 30px;
-  background: url('/@cardstack/boxel/images/icons/expand.svg') center/16px 16px no-repeat;
+  background: url('../../../../../public/@cardstack/boxel/images/icons/expand.svg') center/16px 16px no-repeat;
```
lukemelia, line 19: Does this work? I didn't think that "public" ever found its way into any URL path.

mattmcmanus: Check out the extended commit message fir an explanation https://github.com/cardstack/boxel/commit/0dbff7f2f18cc502110e50ee6e552dacff7cd0a7

mattmcmanus: I can elaborate more if that's not clear enough￼

lukemelia: Makes sense. Thanks!

PR #282, opened 2021-02-19
Refactor/spacing variables 2

Aierie:
CS-200

Commits with values that I'm hesitant to replace are prefixed with `exception:` (thinking we need some exceptions to the scale, but how and when to do this without it creeping up everywhere I'm not sure).


Aierie:
Small issue uncovered with the dashboard component while trying to create a before/after demo:

Screenshot from 2021-02-19 16-27-50

Content that extends outside of the screen will spread out of the background.

in [this commit](https://github.com/cardstack/boxel/pull/282/commits/c863cd19a97d73096e3f6af6c1bc90c1c8ec1751), added background to the body of the dashboard to fix this. Issue with this fix: doesn't fix the fact that the dashboard will not work properly if nested within other components (dashboard body bleeds out the dashboard).

Aierie:
Screenshot from 2021-02-19 16-51-06

Example of the issue mentioned above. This is two dashboards, one on top of another, first one bleeds into the second. Not fixing yet because I'm not sure if dashboard is meant to be used in this context, but do let me know if this is something we need to consider @burieberry, since cards are meant to be used in many contexts. Solution also isn't great because of repeating a color variable.

Aierie:
### DateDivider
Original
date-divider-original

Modified
date-divider-modified

### Menu
Original
menu-original

Modified
menu-modified

### Milestones
Original
milestones-original

Modified
milestones-modified

### Mode indicator
Original
mode-indicator-original

Modified
mode-indicator-modified

### Participant
Original
participant-original

Modified
participant-modified

### Milestone banner
Original
milestone-banner-original

Modified
milestone-banner-modified

burieberry:
> Small issue uncovered with the dashboard component while trying to create a before/after demo:
> 
> Screenshot from 2021-02-19 16-27-50
> 
> Content that extends outside of the screen will spread out of the background.
> 
> in [this commit](https://github.com/cardstack/boxel/pull/282/commits/c863cd19a97d73096e3f6af6c1bc90c1c8ec1751), added background to the body of the dashboard to fix this. Issue with this fix: doesn't fix the fact that the dashboard will not work properly if nested within other components (dashboard body bleeds out the dashboard).

Sorry should've specified this, it's a currently a temporary background for the card pay thread modals. But it's great that you noticed the issues :) I have a branch with better dashboard css that I should do a PR and merge. Why is the content in there repeating though?

Aierie:
> > Small issue uncovered with the dashboard component while trying to create a before/after demo:
> > Screenshot from 2021-02-19 16-27-50
> > Content that extends outside of the screen will spread out of the background.
> > in [this commit](https://github.com/cardstack/boxel/pull/282/commits/c863cd19a97d73096e3f6af6c1bc90c1c8ec1751), added background to the body of the dashboard to fix this. Issue with this fix: doesn't fix the fact that the dashboard will not work properly if nested within other components (dashboard body bleeds out the dashboard).
> 
> Sorry should've specified this, it's a currently a temporary background for the card pay thread modals. But it's great that you noticed the issues :) I have a branch with better dashboard css that I should do a PR and merge. Why is the content in there repeating though?

Ah I added content just to demonstrate that the body would mess up the box spacing

addon/components/boxel/milestones/index.css
```diff
@@ -15,6 +15,7 @@
   content: "";
   width: 1px;
   height: var(--boxel-sp);
+  /* TODO: discuss exceptions for extra small things */
   margin: 2px 0 3px 7px;
```
burieberry, line 19: Maybe we can use `0.1rem` for when it's not really 0 spacing but almost? What do you think @lukemelia ?

burieberry (approved):
These look really good

addon/components/boxel/milestones/index.css
Aierie: Not sure about this, but it looks like the spacing here is more for aesthetics than for visual hierarchy/grouping? I think preserving the current units is important if it's aesthetics, I would leave these units as is.

lukemelia: The context here is that we're drawing a line between milestone items, and the margin is positioning the line. Therefore, our acceptance criteria should be whether the connecting lines look good as we zoom the font size of the page in and out. This is a very exceptional case where margin is not really "margin".

tests/dummy/app/css/components/mode-indicator.css
```diff
@@ -1,56 +0,0 @@
-.mode-indicator {
```
Aierie, line 1: Hey @burieberry. Did a last review before merging, found an issue where we had two different css files for the `mode-indicator` component. Given that design team does reviews based on what's shown on the boxel site, should we replace `addon/components/boxel/mode-indicator/index.css` with this deleted file/merge the two? After deleting this, the `mode-indicator` looks fairly different:

image

Aierie: image

Original

Aierie: There don't appear to be any other components in this PR with a second css file, but will check for the other components as well.

Aierie: Other components don't seem to be affected by removing imports from `tests/dummy/app/css/boxel.js`, based on comparing with github pages.

burieberry: @Aierie The styling for the one you marked "Original" is correct. There should be only one css file and it should be in `index.css` under the addon component files. Can you fix the index.css and remove the other one? Thanks for noticing this, must have been left accidentally.

PR #283, opened 2021-02-19
Refactor/spacing variables 3

Aierie:
CS-200

Locations of things I think should be exceptions to the scale are in commits prefixed with `exception:`.

### DIFF
--- a/addon/components/boxel/selection-control-group/index.css
+++ b/addon/components/boxel/selection-control-group/index.css
@@ -9,20 +9,20 @@
 .selection-control-group__select-all {
   position: relative;
   width: 116px;
-  padding: 11px 27px 10px;
+  padding: var(--boxel-sp-xs) var(--boxel-sp-lg) var(--boxel-sp-xs);
   transition: all var(--boxel-transition);
 }
 
 .selection-control-group__select-all > svg {
-  margin-right: 10px;
+  margin-right: var(--boxel-sp-xs);
 }
 
 .selection-control-group__select-all--selected-items {
   width: 246px;
-  padding-left: 17px;
-  padding-right: 50px;
+  padding-left: var(--boxel-sp-sm);
+  padding-right: var(--boxel-sp-xxl);
 }
 
 .selection-control-group__more-menu {
-  margin: 15px 0;
+  margin: var(--boxel-sp-sm) 0;
 }
--- a/addon/components/boxel/selection-control-group/index.css
+++ b/addon/components/boxel/selection-control-group/index.css
@@ -2,11 +2,6 @@
   display: flex;
 }
 
-.selection-control-group__select-button {
-  /* TODO: discuss exception for margin */
-  margin-right: 1px;
-}
-
 .selection-control-group__select-all {
   position: relative;
   width: 116px;
--- a/addon/components/boxel/selection-control-group/index.css
+++ b/addon/components/boxel/selection-control-group/index.css
@@ -3,6 +3,7 @@
 }
 
 .selection-control-group__select-button {
+  /* TODO: discuss exception for margin */
   margin-right: 1px;
 }
 
--- a/addon/components/boxel/sort-menu/item/index.css
+++ b/addon/components/boxel/sort-menu/item/index.css
@@ -30,8 +30,8 @@
   left: -10px;
 }
 
-/* TODO: discuss exceptions for margin */
 .sort-menu-item__active > span {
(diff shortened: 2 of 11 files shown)


Aierie:
### ParticipantList
Original
participant-list-original

Modified
participant-list-modified

### ParticipantSummary
Original
participants-summary-original

Modified
participants-summary-modified

### SelectionControlGroup
Original
selection-control-group-original

Modified
selection-control-group-modified

### SortMenu
Original
sort-menu-original

Modified
sort-menu-modified

### ThreadMessage
Original
thread-message-original

Modified
thread-message-modified

### WavePlayer
Original
wave-player-original

Modified
wave-player-modified

addon/components/boxel/selection-control-group/index.css
```diff
@@ -3,6 +3,7 @@
 }
 
 .selection-control-group__select-button {
+  /* TODO: discuss exception for margin */
```
lukemelia: @Aierie What is the purpose of this margin? What do you suggest?

addon/components/boxel/sort-menu/item/index.css
```diff
@@ -30,6 +30,7 @@
   left: -10px;
 }
 
+/* TODO: discuss exceptions for margin */
 .sort-menu-item__active > span {
   padding-left: 0.4em;
```
lukemelia, line 35: @Aierie What is the purpose of this margin? What do you suggest?

addon/components/boxel/thread-header/index.css
```diff
@@ -19,6 +19,7 @@
 }
 
 .thread-header__title {
+  /* TODO: discuss exceptions for margin */
```
lukemelia: Same question: What is the purpose of this top margin? What do you suggest?

addon/components/boxel/thread-header/index.css
```diff
@@ -66,6 +67,7 @@
   position: absolute;
   top: 0;
   right: 0;
+  /* TODO: discuss exceptions for padding */
```
lukemelia: @Aierie This one seems more standard. What is its purpose? What do you suggest?

addon/components/boxel/selection-control-group/index.css
Aierie: This seems like a margin placed for precise aesthetic purposes. Considering the document flow is working nicely for this component, I would remove this margin altogether. Pic of margin below (thin orange line):

image

addon/components/boxel/thread-header/index.css
Aierie: There seem to be some visual bugs with this one:
image


If this is how it's supposed to look:
image

I think the purpose of padding here is to provide bulk and make the number of notifications clear and visible. I would keep the padding as is, because of the cramped space and the possibility of overlap:
image
 
Otherwise, I would consider increasing the height of the thread header overall, to provide more space, and use the spacing variables, but wonder if that would require design input.

addon/components/boxel/thread-header/index.css
Aierie: This one seems to be here to create some space so that both the label and title are easier to read, but are close enough that their relationship is clear, they are meant to be read together. I think I would keep this, considering the before v after.

Before applying variable 
image

After rounding 4px to 6px
image

addon/components/boxel/sort-menu/item/index.css
Aierie: This margin provides space between a category to sort by (eg. Title) and its type (A-Z / 1-9) for the currently selected sorting method:
 
image

Because of the use of the `em` unit, it works well across different font sizes. Here it is at 10px (previous picture was 16px):
image

I am not sure if fitting different font sizes is a requirement for this component. If it is, I would keep it as is. If it isn't, I would use the closest spacing variable:
image

addon/components/boxel/thread-header/index.css
Aierie: After discussion with @burieberry , will not apply the spacing variables to these parts of the component.

addon/components/boxel/sort-menu/item/index.css
Aierie: Leaving this spacing as is for now.

addon/components/boxel/selection-control-group/index.css
Aierie: Will remove margin

PR #285, opened 2021-02-23
Remove brand names and logos

burieberry:
master-detail

discrepancies-list

comparison-mode


lukemelia:
Strange test run. It seems that the Lint step passed but then some builds failed with lint errors?

lukemelia:
> Strange test run. It seems that the Lint step passed but then some builds failed with lint errors?

I found a fix here: https://github.com/cardstack/boxel/pull/286

PR #287, opened 2021-02-23
Component tests for Button, SelectButton, Menu, SortMenu, SelectionControlGroup

Aierie:
CS-255
CS-256
CS-257
CS-258
CS-259

Scaffolded todos for tests that I'm going to implement later. Would appreciate feedback @burieberry @lukemelia


addon/components/boxel/button/index.hbs
```diff
@@ -3,7 +3,6 @@
     "boxel-button"
     boxel-button--primary=@primary
     boxel-button--dropdown=@dropdownIcon
-    boxel-button--disabled=@disabled
     boxel-button--collection-style=@collectionStyle
   }}
   disabled={{@disabled}}
```
Aierie, line 8: Made this commit after discussing with Burcu about the case where someone sets a different `disabled` attribute from the `@disabled` argument, and the attribute takes precedence. 
```js
<Boxel::Button @disabled={{false}} disabled={{true}}> Button text </Boxel::Button>
```


The button's styles would be misleading - a disabled button would look normal. With this change, we style directly based on the button's state so what you see is what you get.

@burieberry @lukemelia wondering about the role of the `@disabled` argument after this change? Is there much benefit in exposing `@disabled` now, if it's overwritable by the `disabled` attribute?

Aierie: Not sure how I found this, but apparently there is a use case for making buttons look disabled visually, behave disabled (via pointer events), but not actually have the disabled attribute: 

https://gomakethings.com/dont-disable-buttons-while-submitting-forms-with-ajax/

(mainly raised because it's interesting, don't think this is something we should make easy for users to do)

tests/integration/components/boxel/selection-control-group-test.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Scaffolding tests for SelectionControlGroup component (https://cardstack.github.io/boxel/#/docs?s=Components&ss=%3CBoxel%3A%3ASelectionControlGroup%3E)
+ * path from project root: addon/components/boxel/selection-control-group
+ */
+
+/**
+ * ### Things I'm not sure about
+ * Wondering why we don't toggleSelectAll on the Boxel::Button as well, if its text is select all
```
lukemelia: I've wondered this myself!

tests/integration/components/boxel/selection-control-group-test.js
```diff
@@ -0,0 +1,53 @@
...
+ * - because there's a nested component that wants to be clicked.
+ * - should we still allow people to click it and "Select All" when there are no items selected?
+ * We use ember-basic-dropdown in the DropdownButton component but don't list it in package.json.
+ * It's a dependency of ember-power-select, so it's installed at the moment, I think we should probably specify it as a dependency explicitly so we don't accidentally break things.
```
lukemelia: Agreed

tests/integration/components/boxel/sort-menu-test.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Scaffolding tests for SortMenu component (https://cardstack.github.io/boxel/#/docs?s=Components&ss=%3CBoxel%3A%3ASortMenu%3E)
+ * path from project root: addon/components/boxel/menu
+ */
+
+/**
+ * ### Things I'm not sure about
+ * Wondering why are we allowing the style attribute to be overrideable?
+ * Are styles, like classes, merged?
```
lukemelia: No, they are not.

tests/integration/components/boxel/selection-control-group-test.js
burieberry: Hmm here it's more likely that the `selection-control-group` is a `fieldset`, `select-button` is an `input` of type radio, the text is the label, and the '...' icon that appears is a button.

Aierie: We could implement it that way I guess! Right now it's a bit simpler:

https://github.com/cardstack/boxel/blob/3bca464569843b5e585873ab2aaaadde3d5746c0/addon/components/boxel/selection-control-group/index.hbs#L1-L28

tests/integration/components/boxel/button-test.js
```diff
@@ -0,0 +1,95 @@
...
+
+  test('It renders with the correct html inside', async function (assert) {
+    await render(
+      hbs`<Boxel::Button><span class="test-span">Testing!</span></Boxel::Button>`
+    );
+    assert.dom(`${BUTTON_SELECTOR} span.test-span`).exists();
+  });
+
+  test('It can be clicked and call a callback', async function (assert) {
+    this.set('clicked', false);
```
lukemelia: This could be a simple JS variable. No need to use `set`. i.e. `let clicked = false;` and `clicked = true;` and `assert.ok(clicked);`

tests/integration/components/boxel/menu-test.js
```diff
@@ -0,0 +1,77 @@
...
+
+const MENU_ITEM_SELECTOR = '.menu__item';
+const MENU_SEPARATOR_SELECTOR = '.menu__separator';
+const TEST_MENU_ITEM_TEXT_ATTRIBUTE = 'data-test-menu-item';
+
+module('Integration | Component | Menu', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('It can render a list of menu items that trigger appropriate callbacks when clicked', async function (assert) {
+    this.set('lastClicked', null);
```
lukemelia: Same thing as above. Use simple JS locals here. If you're using a variable in hbs, you should use `set`, if only in JS, use a local.

tests/integration/components/boxel/selection-control-group-test.js
```diff
@@ -0,0 +1,126 @@
...
+
+module('Integration | Component | SelectionControlGroup', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('It can render a state where all items are selected', async function (assert) {
+    this.setProperties({
+      toggled: false,
+      selectedItemCount: 5,
+      toggleSelectAll: () => {
+        this.set('toggled', true);
```
lukemelia: make `toggled` a local and don't use `set`.

lukemelia (changes requested):
Looks good overall, except the issue I left comments about.

Aierie:
Agreed, thanks for spotting the issues with local variables @lukemelia!

lukemelia (approved):

tests/integration/components/boxel/button-test.js
```diff
@@ -0,0 +1,95 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, click } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+const BUTTON_SELECTOR = '.boxel-button';
+const DROPDOWN_BUTTON_SELECTOR = '.boxel-dropdown-button';
```
burieberry: @Aierie  My only suggestion is to use [ember-test-selectors](https://github.com/simplabs/ember-test-selectors) instead of class names. What this means is on the button component's hbs file you would add a `data-test` attribute, such as:

```<button class="boxel-button" data-test-boxel-button ...attributes> ... </button>```

Then when writing the test:
`assert.dom('[data-test-boxel-button]').exists();`

Makes life so much easier, and it means you don't have to worry if you ever have to change the class name in the component.

Here's an example from Cardhost:
- Component: https://github.com/cardstack/cardstack/blob/master/packages/cardhost/app/components/card-size-toggle.hbs
- Test: https://github.com/cardstack/cardstack/blob/master/packages/cardhost/tests/integration/components/card-size-toggle-test.js

Aierie: Ah that seems good! It's nice to have them removed in prod. Will implement that and then merge.

PR #288, opened 2021-02-24
OrgHeader and Dashboard components

burieberry:
Fixes [CS-157](https://linear.app/cardstack/issue/CS-157/dashboard-header-component)

Extracted the Dashboard layout container component and OrgHeader component. Implementation in Card Pay and the Media Registry. No visual changes in media-registry.

dashboard
dashboard-args
org-header
card-pay


lukemelia (approved):
Looks good. I like where you the boundaries are drawn here and the used of named blocks.

Aierie (approved):
Seems to work well, I like the convenient APIs exposed for visual fine-tuning of things like logos, and colors.

nit: flexbox gap doesn't work in safari, so `Boxel::OrgHeader` will not have a gap between the title and other content on that browser if it's crowded.

burieberry:
> Seems to work well, I like the convenient APIs exposed for visual fine-tuning of things like logos, and colors.
> 
> nit: flexbox gap doesn't work in safari, so `Boxel::OrgHeader` will not have a gap between the title and other content on that browser if it's crowded.

Thanks, converted back to grid.

PR #293, opened 2021-03-02
Enhancement/button variations

Aierie:
CS-239

This PR implements variations of the button component and introduces a new API to support the many combinations of buttons that we need to support.

### New API for `Boxel::Button`
**`@size`**  controls the size of the button and its text and icons
accepts values: `small, base, large`
also `collection-style` to support use in `media-registry` without hacks. 

**`@kind`** controls the theming and colors of the icon
accepts values: `primary, secondary-light, secondary-dark`
also `collection-style` to support use in `media-registry` without hacks. 

**`@disabled`** controls whether an icon is disabled or not
accepts values: `boolean`

**`@iconRight`** adds an icon to the right side of the button via `svg-jar` and sets the layout of the button to left-aligned
accepts values: `string`

**`@iconLeft`** adds an icon to the left side of the button via `svg-jar` and sets the layout of the button to left-aligned
accepts values:`string`

### Notes on icon stuff
Opted to let users pass in icons via string because it seemed like it would be the easiest for someone new to the library to get started with. One drawback of this is, our icons are not great used this way at the moment. Icons seem to be cropped to just the parts that have content, which means they are missing whitespace that makes them predictable when used with other icons/contexts. Example below is from material icons (was finding it hard to illustrate examples from our stuff, but `carot-down` is one example which would work better with other icons/text if its whitespace was included).

image

### Notes on `collection-style`
I feel that this should be deprecated, because `collection-style` is not a useful to people who consume boxel as an open source library. Theming would allow us to remove this and still provide correct styles for the example workflow, and I think is good to implement soon. (I like how you separate theme and style in the example you showed yesterday @mattmcmanus!)

### Todo before merge
- Improve freestyle stuff
- Add a way to adjust icon sizing


Aierie:
Light mode

image


Dark mode

image

Icons
image

Aierie:
Do people have any opinions on how icons should be passed in? I've been thinking about:

1. Pass in a string, control icon size via css variable. Color is controlled by the button so it's coordinated with other parts during hover, focus, and other states. I'm assuming that users have an established icon system where say, all 24px-sized icons are meant to work together without much hassle, and the css variable is used as an exception. (current implementation)

2. Provide named blocks for right and left icon. This allows a lot more fine-grained-control, but messes up the api in different ways because you now have to do `<Boxel::Button><:default>button text</:default></Boxel::Button>` when you're using the icons, but when you're not you just do `<Boxel::Button>button text</Boxel::Button>`. It also exposes users to more implementation details than a string would, I think.

addon/components/boxel/selection-control-group/index.hbs
```diff
@@ -1,29 +1,36 @@
...
     @mode={{@mode}}
     @isPartial={{@selectedItemCount}}
     @isSelected={{@isSelected}}
+    {{on 'click' @toggleSelectAll}}
   />
-  <Boxel::Button
+  <div
     class={{cn
-      "selection-control-group__select-all"
+      'selection-control-group__select-all'
```
burieberry, line 19: No need to change these to single quotes. We use double quotes in `hbs` files and single quotes in `js` files. Just a coding style thing. I know it's been inconsistent in boxel, so I've added it to the linter (#294) to make it clearer. I'll wait on merging that PR since it'll create merge conflicts with your PRs.

burieberry:
@Aierie The screenshots look great. I'll take a closer look at the code by the end of my day today, or tomorrow the latest. I didn't get what you meant about the icons, but maybe it'll become clearer when I run it locally. I noticed we're missing a color in the palette.

lukemelia:
> Do people have any opinions on how icons should be passed in? I've been thinking about:
> 
> 1. Pass in a string, control icon size via css variable.

This one. ☝️  Named blocks for this probably provide flexibility that will only come back to cause us problems .

addon/components/boxel/button/size-variants.css
```diff
@@ -0,0 +1,51 @@
...
+  --boxel-button-min-height: 41px;
+}
+
+.boxel-button.boxel-button--size-large {
+  --boxel-button-padding: var(--boxel-sp-xs) var(--boxel-sp-sm);
+  --boxel-button-font: 600 var(--boxel-font);
+  --boxel-button-icon-size: var(--boxel-font-size-sm);
+  --boxel-button-icon-spacing: var(--boxel-sp-xs);
+  --boxel-button-min-height: 46px;
+}
```
lukemelia, line 37: The default `--boxel-button-icon-size` looks like it is the same for all button sizes. Shouldn't it vary to match the button size/font size?

Aierie: I originally made it match font size, but it feels like it's a bit too big on `boxel-button--size-large`

Aierie: For `@size='large'` with `--boxel-buton-icon-size: var(--boxel-font-size)`
image

For `@size='base'` with `--boxel-buton-icon-size: var(--boxel-font-size-sm)`
image

Aierie:
Illustrating the issue with our icons - In use, this is what we want (from prod):

image

But we have to adjust the sizes of the icons manually to get it. The close icon has size 10px width and height while the edit icon has size 15px width and height.

Example with both icons set to 13px width and height:

image

The image below shows all our icons' svg assets with width and height set to 30px:

image

You can see that the majority fill up all the space in the svg. I think whitespace is an important part of the context of an icon and should be preserved so we can avoid these manual adjustments as much as possible.

Aierie:
Following chat with design team, will not implement buttons with icons for now. Design team will add "holding shapes" (with appropriate white space etc) for icons in the future. Waiting on hover and other states from design team. @burieberry @mattmcmanus

addon/components/boxel/button/kind-variants.css
```diff
@@ -0,0 +1,67 @@
...
+ * 
+ */
+
+.boxel-button--kind-primary:not(:disabled) {
+  --boxel-button-color: var(--boxel-highlight);
+  --boxel-button-border: 1px solid var(--boxel-highlight);
+  --boxel-button-text-color: var(--boxel-dark);
+}
+
+.boxel-button--kind-primary:not(:disabled):hover,
```
Aierie, line 21: Some points about these button states:
- these styles are meant for hover, but putting them on for active and focus as well until we get more styles. focus style is important for keyboard navigation - should consider using outline or something different?
- design specified a 10% opacity black overlay. I couldn't think of another way of doing that other than an `::after` pseudo-element, but don't think that should be what we do. People use pseudo-elements for all sorts of stuff and we should keep it available for them without encouraging divitis. So instead I picked the color in XD using the color picker. As a result, icons or other arbitrary content will not be darkened. This becomes more important as we move towards icon buttons, but I think we can cross that bridge when we get to it.

Aierie:
Main difficulties using buttons were theming for custom spacing and colors.

addon/components/boxel/action-chin/index.hbs
```diff
@@ -1,19 +1,21 @@
...
+    'action-chin'
     action-chin--light=this.isMemorializedMode
     action-chin--dark=this.isDataEntryMode
   }}
   ...attributes
 >
   <Boxel::Button
-    {{on "click" (optional @onClickButton)}}
-    @primary={{this.isDataEntryMode}}
+    @size={{'small'}}
```
burieberry: When it's just a string type, you can pass it on as `@size="small"` without the curlies.

addon/components/boxel/button/size-variants.css
```diff
@@ -0,0 +1,43 @@
...
+ * boxel-button--size-@size 
+ * 
+ * Classes for the @size argument can control the following properties:
+ * 
+ * --boxel-button-padding
+ * --boxel-button-min-width
+ * --boxel-button-min-height
+ * --boxel-button-font
+ * --boxel-button-icon-size
+ * --boxel-button-icon-spacing
```
burieberry: We can remove these variables if we're not doing the icons yet.

addon/components/boxel/button/kind-variants.css
burieberry: Not using an actual overlay is a good interpretation and judgement call. There are two good options. If using a solid color, we should add that to the `variables.css`, as something like `boxel-highlight-hover`. The other option is to use a brightness filter (`filter: brightness(0.9);`) in the css. This accomplishes the 10% darkness.

Yes to using outline or adding active and focus styles.

addon/components/boxel/button/size-variants.css
```diff
@@ -0,0 +1,43 @@
...
+ * --boxel-button-letter-spacing
+ * 
+ */
+.boxel-button--size-small {
+  --boxel-button-padding: var(--boxel-sp-xxxs) var(--boxel-sp-sm);
+  --boxel-button-font: 600 var(--boxel-font-sm);
+  --boxel-button-min-height: 30px;
```
burieberry: Let's use `rem`. This is now 32px, so 2rem, and the base button min-height is 2.5rem (40px).

addon/components/boxel/button/index.css
```diff
@@ -1,52 +1,40 @@
...
-  text-align: left;
-}
-
-.boxel-button__more-button {
-  position: absolute;
-  top: 0;
-  right: 8px;
+  --boxel-button-color: #d9d9d9;
+  --boxel-button-border: 1px solid #d9d9d9;
+  --boxel-button-text-color: rgba(0, 0, 0, 0.5);
```
burieberry: In the buttons xd file, these colors are now updated to use the existing color palette .

burieberry (changes requested):
Overall this looks great 👍 . Just needs the styling updates from the new xd file that Gardella has sent.

What do you think about also adding a `full-width` button? It would only change the width to be 100%.

addon/components/boxel/action-chin/index.hbs
Aierie: yep thanks, done

Aierie:
note - with latest changes, `@size` argument accepts `small`, `base`, `tall`, and `touch` to more accurately reflect the specs for padding.

`base` is now 32px height button as this is the most widely used button in the ui. the old `base` is now `tall`

Aierie:
also, focus styles are applied as outlines but active styles are still the same as hover at the moment.

Aierie:
Design team has also specified icon buttons + their states. I believe this should be in a separate PR, and should be added to the existing IconButton component

burieberry (approved):

PR #295, opened 2021-03-04
Add Boxel::Field and Boxel::Input components

burieberry:
field

input


addon/components/boxel/input/index.css
```diff
@@ -0,0 +1,28 @@
...
+  outline-color: var(--boxel-highlight);
+}
+
+.boxel-input:disabled {
+  background-color: var(--boxel-light-100);
+  border-color: var(--boxel-purple-300);
+  color: var(--boxel-purple-300);
+}
+
+/* TODO: required and disabled styles */
```
lukemelia, line 28: What's the status of these?

addon/components/boxel/input/index.css
```diff
@@ -0,0 +1,28 @@
+.boxel-input {
+  --boxel-input-height: 2.5rem; /* 40px */
+
+  width: 100%;
+  min-height: var(--boxel-input-height);
+  padding: var(--boxel-sp-xxs) 0 var(--boxel-sp-xxs) var(--boxel-sp);
+  border: 1px solid var(--boxel-purple-300);
+  border-radius: 5px;
```
lukemelia: Consider using ` --boxel-border-radius-sm`

addon/components/boxel/input/index.js
```diff
@@ -0,0 +1,14 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+export default class Input extends Component {
+  @tracked value;
```
lukemelia: It doesn't look like this `value` is used by anything. The template using `@value`

package.json
```diff
@@ -40,6 +40,7 @@
     "ember-cli-string-helpers": "^5.0.0",
     "ember-composable-helpers": "^4.4.1",
     "ember-concurrency-decorators": "^2.0.3",
+    "ember-css-element-queries": "^1.0.0",
```
lukemelia: This addon's repository has been archived (https://github.com/simplabs/ember-css-element-queries). Should we consider a different implementation?

lukemelia (changes requested):
General approach looks great. Left a few questions/suggestions.

package.json
burieberry: Yeah, I wonder why they archived it. Maybe `css-element-queries` (https://github.com/marcj/css-element-queries) now works for ember out-of-the-box. I was planning to try but forgot about it.

burieberry: Update on this: It didn't work out-of-the-box. There's this note on the readme (https://github.com/marcj/css-element-queries#module-loader), but I'm not sure where it needs to be added.

burieberry:
@lukemelia By the way, dependency lint started failing on the `ubuntu-latest` and `ember-beta` steps for all our PRs. We can add a config file for it, but it's strange that it's not failing for the other tests or that it just started failing.

package.json
lukemelia: I saw that you removed it. If you do want to use it, I'm happy to help figure out how to get the css-element-queries module to work.

addon/components/boxel/input/index.hbs
```diff
@@ -1,8 +1,9 @@
 <Input
   class="boxel-input"
   @id={{@id}}
-  @value={{@value}}
+  @value={{this.value}}
```
lukemelia: Why don't we want to just rely on data-down/actions-up for this and require @onInput to update the passed in @value?

package.json
burieberry: Ok we can make a separate ticket for this

addon/components/boxel/input/index.hbs
burieberry: Yes, that was the whole point. I got tired.

lukemelia (approved):

burieberry:
@lukemelia FYI, I don't know if you've seen it but I've also added a TextField component which is a pre-packaged version of Boxel::Field wrapper with text type Boxel::Input. I'm thinking of doing the same for other form field types. If having that wrapper Field component doesn't work out or if it becomes unnecessary we can go back to not doing it that way.

burieberry:
Ah also I'll double check the disabled etc state styling with Gardella during design review, just so that I don't delay merging this any longer.

PR #296, opened 2021-03-05
Card catalog tray

Aierie:
CS-229
CS-230

WIP at the moment, needs cleanup.

### DIFF
--- a/addon/components/boxel/card-catalog-tray-item/index.css
+++ b/addon/components/boxel/card-catalog-tray-item/index.css
@@ -23,7 +23,7 @@
 }
 
 .boxel-card-catalog-tray-item--dragged-item {
-  --boxel-card-catalog-tray-item-border: 1px solid rgba(39,35,48, 0.15);
+  --boxel-card-catalog-tray-item-border: 1px solid rgba(39,35,48, 0.15); /* #272330 with 15% opacity */
   filter:  drop-shadow(0 15px 30px rgba(0, 0, 0, 0.5));
 }
 
--- a/addon/components/boxel/card-catalog-tray-item/index.css
+++ b/addon/components/boxel/card-catalog-tray-item/index.css
@@ -1,0 +1,78 @@
+.boxel-card-catalog-tray-item {
+  --boxel-card-catalog-tray-item-title-color: var(--boxel-dark);
+  --boxel-card-catalog-tray-item-description-color: var(--boxel-purple-800);
+  --boxel-card-catalog-tray-item-icon-color: var(--boxel-teal);
+  --boxel-card-catalog-tray-item-background-color: var(--boxel-purple-100);
+  --boxel-card-catalog-tray-item-border: 1px solid #272330;
+  width: 300px;
+  min-height: 80px;
+  padding: var(--boxel-sp-sm) var(--boxel-sp-xs) var(--boxel-sp-sm) var(--boxel-sp-lg);
+  position: relative;
+  border: var(--boxel-card-catalog-tray-item-border);
+  border-radius: var(--boxel-border-radius);
+  background-color: var(--boxel-card-catalog-tray-item-background-color);
+
+  display: grid;
+  grid-template-rows: 20px 30px;
+  grid-template-columns: 20px 1fr 8px;
+  gap: 0px var(--boxel-sp);
+  grid-template-areas:
+    'icon title drag'
+    'icon description drag';
+  align-items: center;
+}
+
+.boxel-card-catalog-tray-item--dragged-item {
+  --boxel-card-catalog-tray-item-border: 1px solid rgba(39,35,48, 0.15);
+  filter:  drop-shadow(0 15px 30px rgba(0, 0, 0, 0.5));
+}
+
+.boxel-card-catalog-tray-item--used,.boxel-card-catalog-tray-item--dragged-in-tray {
+  --boxel-card-catalog-tray-item-title-color: var(--boxel-purple-400);
+  --boxel-card-catalog-tray-item-description-color: var(--boxel-purple-400);
+  --boxel-card-catalog-tray-item-icon-color: var(--boxel-purple-400);
+  --boxel-card-catalog-tray-item-background-color: var(--boxel-light-600);
+  --boxel-card-catalog-tray-item-border: 1px solid #272330;
+}
+
+.boxel-card-catalog-tray-item__icon {
+  grid-area: icon;
+  width: 20px;
+  height: 20px;
+  --icon-color: var(--boxel-card-catalog-tray-item-icon-color);
+}
+
+.boxel-card-catalog-tray-item__title {
+  grid-area: title;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  font: 600 var(--boxel-font-sm);
+  letter-spacing: var(--boxel-lsp);
+  color: var(--boxel-card-catalog-tray-item-title-color);
+}
+
+.boxel-card-catalog-tray-item__description {
+  grid-area: description;
+  height: 30px;
+  margin: 0;
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 2;
+  overflow: hidden;
+  font: var(--boxel-font-xs);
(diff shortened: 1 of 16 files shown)


addon/components/boxel/searchbox/index.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="boxel-searchbox" >
```
Aierie: @burieberry I've implemented the searchbox styling mostly with elements outside the searchbox itself (some parts like the placeholder still need to be styled with pseudoclasses that interact with shadow dom). An alternative I was considering but ultimately decided not to do was to put most of the styles in the shadow dom of the input component. Decided against it because it feels like a more fragile solution. Do you have any thoughts on this?

burieberry: I've considered shadow dom for boxel, as well, and decided against it. It also requires writing styles inline, and that's annoying.

addon/components/boxel/searchbox/index.hbs
```diff
@@ -0,0 +1,17 @@
...
+      @id={{@id}}
+      @value={{@value}}
+      {{on "input" (optional @onInput)}}
+      {{on "change" (optional @onChange)}}
+      disabled={{@disabled}}
+      placeholder="Search" 
+    />
+    <button class="boxel-searchbox__clear-button" {{on "click" (optional @onClear)}}>
+      {{svg-jar "close" width="100%" height="100%"}}
+    </button>
```
burieberry: I'm not sure if this X is for the search box. It might be for closing the catalog tray.

Aierie: Yep, there's one for the catalog tray already; this is specifically for the search box.

Aierie:
Some things I'm not sure about @burieberry :
- where are the light theme and dark theme for `Searchbox` and `CardCatalogTrayItem`?
- can we delay adding themes until we formalize a theming api for Boxel?
- should `CardCatalogTrayItem` and `DropTarget` be placed in Cardhost instead of Boxel?

Also note:
- `CardCatalogTray` is not a component because unsure how we should implement logic. Instead, there is an example in the `CardCatalogTrayItem` documentation
- drag and drop is just to show how the `dragged-item` and `dragged-in-tray` styles for  are used.

Aierie:
Using the card catalog tray

https://user-images.githubusercontent.com/39579264/110478278-26cefa00-811f-11eb-83f8-1ac05c84cbf9.mov

Drop target
image

addon/components/boxel/card-catalog-tray-item/index.css
```diff
@@ -0,0 +1,78 @@
+.boxel-card-catalog-tray-item {
+  --boxel-card-catalog-tray-item-title-color: var(--boxel-dark);
+  --boxel-card-catalog-tray-item-description-color: var(--boxel-purple-800);
+  --boxel-card-catalog-tray-item-icon-color: var(--boxel-teal);
+  --boxel-card-catalog-tray-item-background-color: var(--boxel-purple-100);
+  --boxel-card-catalog-tray-item-border: 1px solid #272330;
```
burieberry: Let's add this color to `variables.css` as `--boxel-purple-900`. We actually used this color a lot in cardhost, and now we'll be using it in this repo.

addon/components/boxel/card-catalog-tray-item/usage.js
```diff
@@ -0,0 +1,74 @@
...
+      state: null,
+      icon: 'gear',
+      title: 'Second item',
+      description: 'Unused item',
+    },
+    {
+      state: null,
+      icon: 'eye',
+      title: 'Third item',
+      description: 'Unused item',
```
burieberry, line 49: Let's use a longer copy from the design file for one of them, so we can see an example with 2 lines of description

addon/components/boxel/searchbox/index.hbs
```diff
@@ -0,0 +1,23 @@
+<div class="boxel-searchbox" ...attributes>
+    {{svg-jar "search" class="boxel-searchbox__search-icon"}}
+    {{#if (and @id @label)}}
+      <label for={{@id}} class="boxel-searchbox__input-label">{{@label}}</label>
```
burieberry, line 4: I think the `@label` value here should be the same as the placeholder value. The hidden label is only here because placeholder by itself is not enough for accessibility, so they're sort of serving the same purpose.

addon/components/boxel/searchbox/index.hbs
```diff
@@ -0,0 +1,23 @@
...
+    {{/if}}
+    {{!-- TODO: review accessibility of type=text vs type=search --}}
```
burieberry, line 6: 👍

public/images/icons/gripper.svg
```diff
@@ -0,0 +1,20 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="9" height="24">
```
burieberry: Can you pass this through ImageOptim so the extra stuff is removed?

burieberry (approved):
Looks good. After you're done with changes and this PR is merged, don't forget to create a ticket for gardella so that he can do a design review.

addon/components/boxel/searchbox/index.hbs
Aierie: Label can sometimes be different from placeholder I think. Visual hierarchy can support simpler/briefer copywriting, which the label can compensate for.

image


I think making a `@placeholder` argument is good though!

public/images/icons/gripper.svg
Aierie: I went through SVGO (https://jakearchibald.github.io/svgomg/). I didn't realise that svg xmlns was actually an optional thing for inline so I didn't remove it, will remove now.

addon/components/boxel/card-catalog-tray-item/index.css
```diff
@@ -0,0 +1,78 @@
...
+  grid-area: drag;
+  margin-left: auto;
+  cursor: grab;
+}
+
+.boxel-card-catalog-tray-item__used-icon {
+  position: absolute;
+  right: var(--boxel-sp-xxxs);
+  top: var(--boxel-sp-xxxs);
+}
```
lukemelia, line 82: Lots of `px` values here. Seems like most should be converted to `rem`?

addon/components/boxel/card-catalog-tray-item/usage.css
```diff
@@ -0,0 +1,79 @@
+/* a dummy item to do drag and drop */
+.very-far-away {
+  position: absolute;
+  top: -1000px;
+}
+
+/* card catalog tray usage */
+.boxel-card-catalog-tray-example {
+  /* background-image: url('../@cardstack/boxel/images/icons/close.svg'); */
```
lukemelia: Should this be removed?

lukemelia (commented):
Nice work. Left some notes inline.

I don't think the "used" state makes sense in real life. You can add more than one instance of a card so when do we think "used" would come into play?

The tray's height probably should not collapse when no cards match the filter text.

The "clear" X next to the "close" X is confusing. (non-designer's design opinion)

It looks like the search box has a blue outline AND a white outline when focused. Overkill? (non-designer's design opinion)

Aierie:
> I don't think the "used" state makes sense in real life. You can add more than one instance of a card so when do we think "used" would come into play?

> The "clear" X next to the "close" X is confusing. (non-designer's design opinion)

Thanks! I'll bring these up with design when I ask for review. I think the searchbox close button should have less emphasis. There was a close button as part of the design but it was a hidden layer with a black color, I figured would be easier to let design review it as part of the full component.

> It looks like the search box has a blue outline AND a white outline when focused. Overkill? (non-designer's design opinion)

This is actually Chrome's default focus style on a dark background, I don't like it either. Design might do some specifications for focus styles to implement globally in Boxel, should resolve this issue.

@lukemelia

addon/components/boxel/card-catalog-tray-item/index.css
Aierie: Not sure how they scale with `rem`. Am testing out.

Aierie: Adjusted to use rem

PR #297, opened 2021-03-09
Workaround broken release of workerpool

lukemelia:


lukemelia:
This didn't work. We're going to wait for the community to resolve this. Shouldn't be long.

PR #299, opened 2021-03-11
Prefix component CSS classes with "boxel-"

lukemelia:
In order for the Boxel Design System Ember Addon to be eventually adopted by third-parties, it needs to avoid likely clashes with application CSS. This commit makes sure all our CSS for Boxel components are prefixed with "boxel-".


burieberry:
@lukemelia I merged this PR since it's so big, but noticed two places where styling is broken. We can fix those in later PRs.
wave-player
Screen Shot 2021-03-11 at 10 22 33 AM

breadcrumbs
vs.
breadcrumbs-discrepancy

lukemelia:
>@lukemelia I merged this PR since it's so big, but noticed two places where styling is broken. We can fix those in later PRs.

Ah, I looked in the addon folder but didn't think to look in the scenarios.

PR #300, opened 2021-03-11
Ci changes

Aierie:
Changes to github actions for the deployment previews.


Aierie:
@lukemelia @mattmcmanus  is there a reason why we run CI/CD on every push? Am thinking of having the entire preview deployment workflow be separate from the deploy one. Both would run the same tests, but trigger upon different events. 

The preview deployment workflow would trigger upon `pull_request` events with `master` as the base. By default, this includes `opened`, `reopened`, and `synchronize`, so new, reopened, and updated PRs that target `master` would go through the necessary tests.

The production deployment workflow would remain the same, but trigger only upon `push` events on `master`.

*Note: the current implementation of preview deployment is WIP

lukemelia:
@Aierie That could work. I would only be concerned about keeping lint + test steps config in sync across two workflows.

Aierie:
Added on to current workflow but changed triggers from every push to:

1. Updates to `master`
2. Opening, reopening, and updating pull requests that have `master` as base

PR #301, opened 2021-03-12
Card Space setup

burieberry:
[CS-233](https://linear.app/cardstack/issue/CS-233/set-up-routing-and-templates-for-cardspace-scenario)

Changes:
- Added `/scenarios/cardstack` route and nested the Card Pay and the new Card Space routes under it
- Mock login service
- Edges service and logic around when to display the left edge and the login button
(Left edge is always present on `cardstack` and `cardstack.card-space` routes. It only appears after login on user space pages.)
- Currently using ember-toolbars for the left-edge
- Styling changes for Dashboard component and dark-themed version


tests/dummy/app/controllers/cardstack/card-space/new.js
```diff
@@ -0,0 +1,18 @@
...
+export default class CardSpaceNewController extends Controller {
+  @service('edges') edges;
+  @service('cardstack-session') cardstackSession;
+
+  @action updateEdges() {
+    if (this.cardstackSession.isAuthenticated) {
+      this.edges.showLeftEdge();
+    } else {
+      this.edges.hideLeftEdge();
+    }
```
lukemelia: Maybe the edges service should offer `updateDisplayLeftEdge(isDisplayed:boolean)` so that this code can be simpler.

lukemelia (approved):
Looks like a nice clean start!

PR #306, opened 2021-03-16
Add Card Container and Layout Container components

burieberry:
Fixes [CS-368](https://linear.app/cardstack/issue/CS-368/update-cardheader-component)
Fixes [CS-371](https://linear.app/cardstack/issue/CS-371/card-container-component)

Changes:
- API and styling updates for the Boxel::Header component
- Added Boxel::CardContainer component
- Used the CardContainer and Header components for ActionContainer and LayoutContainer
- Added tests for each

header

card-container

action-container

layout-container


lukemelia (approved):
👏  These components are key! Nice work.

Aierie:
Overall seems good! Wondering about a few things:
1. How are we going to add the dropdown menu for the '...'? 
image

2. Button in the `LayoutContainer` header has odd focus states. We do need something to signal interactivity and this does do it, but this feels a bit odd. I think this needs some specification from design on the tabbing behaviour for a proper solution though, so would leave this as is.
image

3. If we change the word 'Layout', `LayoutContainer` can also be used for non-layout cards, so it feels a bit awkward if we end up making another one for non-layout cards which have the same affordances in the top right corner that allows a dynamic card name. Is it okay to allow people to pass the name of the card in via an argument?
image
image

4. There are a couple of states for cards. Wondering whether it is good to have a component that implements them here, or if it is best left to the actual usage via `CardContainer`.
image
image
image
image

burieberry:
> Overall seems good! Wondering about a few things:
> 
> 1. How are we going to add the dropdown menu for the '...'?
>    image
> 2. Button in the `LayoutContainer` header has odd focus states. We do need something to signal interactivity and this does do it, but this feels a bit odd. I think this needs some specification from design on the tabbing behaviour for a proper solution though.
>    image
> 3. If we change the word 'Layout', `LayoutContainer` can also be used for non-layout cards, so it feels a bit awkward if we end up making another one for non-layout cards which have the same affordances in the top right corner that allows a dynamic card name. Is it okay to allow people to pass the name of the card in via an argument?
>    image
>    image
> 4. There are a couple of states for cards. Wondering whether it is good to have a component that implements them here, or if it is best left to the actual usage via `CardContainer`.
>    image
>    image
>    image
>    image

Here's what I'm thinking:

1- We can use the Boxel DropdownButton component for the menu. I haven't implemented it in this PR.

2- Yes, the focus tab styling is ugly and we should do a discuss with the design team. So far we're only using that in Card Space though, so maybe it can wait until we go back to working on Card Space.

3 and 4- Good point. I'm thinking of those cards as "embedded layout container"s. So we would use the current LayoutContainer for the outer container, and make an embedded version for the other cards. The view mode for those cards look different (they keep their boundaries and rounded corners, unlike the non-embedded one.) and they have the different states you've mentioned. More detailed explanation [here](https://docs.google.com/document/d/1xHLsur-dnGUbhLuqMU9kQHLr-TzQmuO6clTUSMYl6OI/edit?usp=sharing), even though I haven't updated it in a while. Will go back to implementing this during Card Space work.

I'll update the PR to add these as To-Do items.

Aierie (approved):
Cool, I think keeping changes minimal till we use the components is good then!

burieberry:
@lukemelia experiencing random test fail again, do you know why ember-beta is failing?

lukemelia:
> @lukemelia experiencing random test fail again, do you know why ember-beta is failing?

@burieberry I think we need to upgrade to embroider 0.37.0 to get https://github.com/embroider-build/embroider/commit/5bd5507cda84e28ff37df2a14040789a71e6276c /cc @Aierie

Aierie:
@lukemelia Upgrading to embroider 0.37.0 creates new errors that cause tests to fail both on our current version of ember, and `ember-beta`. Might need some help with debugging them

lukemelia:
@Aierie phone a friend

PR #308, opened 2021-03-18
Don't run preview deployments from external PRs

Aierie:


.github/workflows/test-and-deploy.yml
```diff
@@ -229,7 +229,7 @@ jobs:
   deploy-preview:
     name: Deploy a preview to S3
     runs-on: ubuntu-latest
-    if: github.head_ref
+    if: github.event.pull_request.head.repo.full_name == github.repository && github.head_ref
```
lukemelia, line 234: Please add a comment describing what this is checking since it may not be obvious to a future reader

lukemelia (changes requested):

.github/workflows/test-and-deploy.yml
Aierie: Done!

lukemelia (approved):

PR #309, opened 2021-03-18
Focus styles

Aierie:
CS-470

Use simple outline for focus styles.


burieberry (approved):
The preview site feature came very handy!

I like the `:focus:not(:focus-visible)` trick. That's awesome. Makes the UI look much cleaner, yet accessible. And I like that the focus styling is controlled globally. 👍

PR #310, opened 2021-03-19
Broken style fixes

burieberry:
- remove some unused assets
- focus styling
- dropdown button and menu fixes
- fix missing css class names


lukemelia:
I don't think we need to / should remove the audio file. My understanding is that @Aierie replaced it with content that is in public domain.

Aierie:
Yep! Audio file should be this one: https://pool.publicdomainproject.org/index.php/Busy-bee-d47-3259

I think the canvas rendering based on audio is an important part of the styling, and being able to listen to it without having to provide my own audio file would be nice too, so would also prefer keeping an audio file.

Aierie:
Thanks for catching more cases for focus styles!

Aierie (approved):

PR #311, opened 2021-03-22
Extract boxel modal

Aierie:
CS-487


burieberry (approved):
Good improvement over the existing modal component, especially with the outside click. I saw there are a couple of `@class` artifacts left over, otherwise it's ready to go.

PR #312, opened 2021-03-24
Org landing page

Aierie:
CS-461

image

image


tests/dummy/app/templates/cardstack/index.hbs
```diff
@@ -3,28 +3,53 @@
...
-  </:header>
   <:body>
+  <div class="cardstack-container">
+    <div class="cardstack-section-header">
+    <h2 class="cardstack-section-header__title" data-test-cardstack-dashboard-header>
+      Cardstack
+    </h2>
+    {{#unless this.cardstackSession.isAuthenticated}}
+    <Boxel::Button
+      {{!-- @kind={{if this.edges.darkTheme "secondary-dark" "secondary-light"}} --}}
```
burieberry: You can remove the commented out code on this template

public/images/icons/card-catalog-logo.svg
```diff
@@ -0,0 +1,14 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 70 70">
```
burieberry, line 1: Don't forget to pass the images through ImageOptim. Cleans them up and reduces files sizes nicely.

tests/dummy/app/templates/cardstack/index.hbs
burieberry: Also since we don't need the `darkTheme` stuff anymore, you can remove it from the edges service. It was annoying to have it there anyway. You can remove the test for that, too.

burieberry (changes requested):
Looks good. Made a couple small comments. We should also get the styling for smaller viewports. That can be a follow-up PR if you want to merge this in first.

tests/dummy/app/templates/cardstack/index.hbs
Aierie: Oops. Thanks for catching that one. Will do!

public/images/icons/card-catalog-logo.svg
Aierie: Yep, I did optimize the svgs. I kept `xmlns` because we need it if we're using svgs in a way besides inline, which I did here via `img`. I used `img` because the svgs here don't exactly need to be svgs (unlike icons), they could be any other image type in the future, and it's easier to replace images if all we need to do is replace a url.

Aierie:
> Looks good. Made a couple small comments. We should also get the styling for smaller viewports. That can be a follow-up PR if you want to merge this in first.

re: merging first, I feel like this is better merged after #315, so in the meantime will talk to design about how they would like it to look on smaller screens. There are also designs for the mobile versions of the org-card that I can work on in the meantime.

edit: 
1. design team will take some time specifying designs for other viewport sizes, so we should go ahead with this first.
2. Before that happens, the "CARDSTACK" text on the top banner should be replaced with an svg.

burieberry (approved):

tests/dummy/app/css/templates/cardstack.css
```diff
@@ -1,5 +1,83 @@
...
-  top: var(--boxel-sp);
-  right: var(--boxel-sp);
+.cardstack-container {
+  --card-width: 33.125rem;
+  --horizontal-card-gap: var(--boxel-sp-lg);
+
+  width: 100%;
+  max-width: calc(var(--card-width) * 2 + var(--horizontal-card-gap));
+  margin: auto;
+  padding-bottom: 9.375rem;
```
burieberry: We can make the layout more flexible even without designs for smaller screens. I'd add a `padding: var(--boxel-sp-lg)` instead of the padding-bottom here. And then add `margin-bottom: var(--boxel-sp-lg);` to each of its immediate children.

tests/dummy/app/css/templates/cardstack.css
```diff
@@ -1,5 +1,83 @@
...
 }
+
+.cardstack-section-header {
+  margin: var(--boxel-sp-lg) 0;
+  margin: var(--boxel-sp-lg) 0;
```
burieberry: Since the padding will take care of the top spacing, this can just be `margin-bottom: var(--boxel-sp-lg);`

tests/dummy/app/css/templates/cardstack.css
```diff
@@ -1,5 +1,83 @@
...
+}
+
+.cardstack-section-header__title {
+  font: 600 var(--boxel-font-lg);
+  letter-spacing: 0;
+}
+
+.cardstack-landing-hero {
+  width: 100%;
+  height: 18.75rem;
```
burieberry: Make this `min-height` instead of a fixed height, so that when the page shrinks, the text content doesn't get cropped, and something like a `padding: var(--boxel-sp);` so that the container maintains some spacing between its boundary and the content.

burieberry (commented):
I was about to fix the conflict and merge this, but noticed a couple of things we can do to make the layout more flexible.

Aierie:
@burieberry I've applied changes to tidy up the layout. There's many ways we can polish this, but I think we can refine that when we go through review from the design team, since a lot of things are still in flux. The main thing I would wait for is the mobile design specs.

See these commits for relevant changes:

[4bf0e2a](https://github.com/cardstack/boxel/pull/312/commits/4bf0e2a8620ccc341ebad5f6363d0ed01c507348)
[1db0c48](https://github.com/cardstack/boxel/pull/312/commits/1db0c48bd8c378982b5a8459f06627d38e8c1329)

PR #316, opened 2021-03-25
Expandable banner component

Aierie:
CS-414
image

image


Aierie:
Some rationale on the design of this component:

### Why details/summary elements?
It requires no JS and is decently accessible as a [disclosure widget](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure) without having to specify a lot of its behavior. 

### Why arguments for `@icon` and `@summary` instead of a named block?
- It's probably good to restrict how users (and us) customize icon placement and size here, optimistically. The example in Freestyle has the payment icon with a slightly different vertical placement from that specified in design, but I think we should solve it at a higher level. We should be adding holding shapes to the icons to solve visual deviations from design - this is also an issue that affects adding icons to our existing buttons. 
- You can't add semantic html to a summary element (a h1 is no different from a span inside one), so not allowing users to do this is better. If we are using this as something like a heading, we should consider a different implementation that allows for semantics.

For more information:
https://adrianroselli.com/2020/05/disclosure-widgets.html
https://adrianroselli.com/2019/04/details-summary-are-not-insert-control-here.html

Aierie:
Some concerns about behaviour, will check with design team:
- Is clicking anywhere on the closed element supposed to expand it, or only the hide/show marker?
- Ditto for the expanded element
Mainly concerned about this because the expanded state doesn't delineate the summary element, so it feels like it's just the hide/show marker that should be a trigger.

*Talked to design, design mentioned the entire top bar should be a trigger, both in the closed and expanded states. @burieberry I still feel like the delineation of the top bar isn't clear and can be annoying, but not sure if it's just me. If you feel the same, do let me know.

lukemelia (approved):
Helpful comments in the CSS 👍

burieberry (approved):
Looks good, I think it's nicer to have the whole header clickable whenever we have an expandable component like this, so I agree with the design choice. An additional improvement would be adding css transitions to the expansion and closing, and rotating the marker during transition from one state to the other.

PR #318, opened 2021-03-26
Thread components usage documentation and updates

burieberry:
headers
message
thread
thread-workflow


lukemelia (commented):
Good improvements. In the preview, I notice that we are animating on initial render but it seems like it would be better to animate only when a thread item is being added after initial render.

burieberry:
> Good improvements. In the preview, I notice that we are animating on initial render but it seems like it would be better to animate only when a thread item is being added after initial render.

I took the animations out of the boxel component for now (all in the dummy app now) and made the initial render conditional on whether that step has been completed.

PR #319, opened 2021-03-29
Fix deployment comment

Aierie:
CS-506


Aierie:
In the future, could aim to have comment include informative messages like "Latest deployment failed" by updating it using https://github.com/peter-evans/create-or-update-comment

lukemelia (commented):
Overall, this change looks good, but can we just for for the link `[Link to preview](${{ env.S3_PREVIEW_WEBSITE_URL }}/${{ github.head_ref }})` instead of doing the "comment identifier" approach, which is kind of noisy. When we discussed it, I didn't realize the comment identifier would be visible.

Aierie:
Using the link as a comment identifier does make sense, I'm not sure why it didn't occur to me. Will make the change.

lukemelia (approved):

PR #320, opened 2021-03-30
Enhancement/button links

Aierie:
CS-508

### DIFF
--- a/addon/components/boxel/button/index.hbs
+++ b/addon/components/boxel/button/index.hbs
@@ -1,3 +1,17 @@
+{{#if @href}}
+<a
+  class={{cn
+    "boxel-button"
+    (concat "boxel-button--size-" (or @size this.defaultSize))
+    (concat "boxel-button--kind-" (or @kind this.defaultKind))
+  }} 
+  href={{unless @disabled @href}}
+  data-test-boxel-button 
+  ...attributes
+>
+  {{yield}}
+</a>
+{{else}}
 <button
   class={{cn
     "boxel-button"
@@ -10,3 +24,4 @@
 >
   {{yield}}
 </button>
+{{/if}}
--- a/addon/components/boxel/button/index.css
+++ b/addon/components/boxel/button/index.css
@@ -28,8 +28,19 @@
   letter-spacing: var(--boxel-button-letter-spacing, var(--boxel-lsp-lg));
 }
 
-.boxel-button:disabled {
+/* select disabled buttons and links that don't have href */
+.boxel-button:disabled, a.boxel-button:not([href]),a.boxel-button[href=""] {
   --boxel-button-color: var(--boxel-purple-300);
   --boxel-button-border: 1px solid var(--boxel-purple-300);
   --boxel-button-text-color: var(--boxel-purple-400);
+}
+
+/* the a element does not have a disabled attribute. Clicking will still trigger event listeners */
+a.boxel-button:not([href]),a.boxel-button[href=""] {
+  pointer-events: none;
+}
+
+/* overwrite the global style for links in global.css */
+a.boxel-button:hover {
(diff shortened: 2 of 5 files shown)


Aierie:
Context - we have some CTA-ish visual elements whose primary purpose is to send people to a particular page (either within or outside our apps). The best way to do this is to use the anchor element.

These CTAs have visual styling similar to a button, so we can add an anchor that looks like a button without duplication of styles by putting the anchor within the `Boxel::Button` component and using the anchor tag if a `@href` is passed to the component. However, the disabled state is a special case, because anchors do not accept the `disabled` html attribute. 

A few important points about buttons which are using the anchor element:
1. When disabled, they are placeholder links (without a `href` attribute): https://www.w3.org/TR/2011/WD-html5-author-20110809/the-a-element.html. We do this programmatically - omitting the `href` attribute when the `@disabled` argument is true. 
2. We should be wary of the attributes that we provide a disabled link, because they can mess up screen readers: https://www.scottohara.me/note/2019/07/17/placeholder-link.html. Post is from 2019, but it still seems quite relevant, you can try out the page that he shared by (on Mac) clicking Command + F5 to bring up Voiceover.
3. Programmatic click events will still trigger even with the `pointer-events:none` style (and when the element can be navigated to with the keyboard, the Enter key will still trigger events). This is mainly relevant for screen readers, and our tests - since tests simulate clicks by dispatching events to elements.

Aierie:
An alternative approach is to follow Carbon Design System's choice of rendering a button instead of an anchor for the disabled state. This way we need to worry less about the screen reader and programmatic access breaking a "disabled" state, but I'm not sure how this behaves with screen readers.

See: https://github.com/carbon-design-system/carbon/blob/7a80e679a9d6c455614d427c121bd8cce03d728b/packages/react/src/components/Button/Button.js#L142-L168

addon/components/boxel/button/index.hbs
```diff
@@ -1,3 +1,17 @@
...
+    "boxel-button"
+    (concat "boxel-button--size-" (or @size this.defaultSize))
+    (concat "boxel-button--kind-" (or @kind this.defaultKind))
+  }} 
+  href={{unless @disabled @href}}
+  data-test-boxel-button 
+  ...attributes
+>
+  {{yield}}
+</a>
```
burieberry, line 13: Let's prioritize `LinkTo` usage. Our button-looking links are going to require it.

Aierie: I've found ways around that, like https://github.com/buschtoens/ember-link#link-helper. 

But if we want to link to another part of Cardstack, eg. the marketing page from the app, is there a workaround for `LinkTo`?

lukemelia (approved):

addon/components/boxel/button/index.hbs
Aierie: Alternatively, we can have two arguments - `@href` or `@route`, and render `a` or `LinkTo` depending on which one is given.

burieberry: > Alternatively, we can have two arguments - `@href` or `@route`, and render `a` or `LinkTo` depending on which one is given.

Yes, this is what I was thinking. We seem to use text links for external links, but we can implement just in case.

Aierie: Was looking at the `LinkTo` component and thinking about the API. The `LinkTo` component does more than I initially thought. How much of its functionality do you think we will need @burieberry ?

https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/LinkTo?anchor=LinkTo

Aierie: Or rather, how much should we expose on the button component?

Aierie:
Closed in favor of #325

PR #322, opened 2021-03-31
suggested next steps box component

Aierie:
CS-161

image


tests/dummy/app/components/card-pay/suggested-next-steps-box/usage.hbs
```diff
@@ -0,0 +1,29 @@
+<Freestyle::Usage @name="SuggestedNextStepsBox">
+  <:example>
+    <CardPay::SuggestedNextStepsBox class="usage-suggested-next-steps-box">
+      <:body as |BodyButton|>
+        {{#each this.nextSteps as |item|}}
+          <BodyButton @href={{item.href}}>
```
Aierie: using href here first, will only work if #320 is merged.

lukemelia (approved):

burieberry (commented):
It looks good, but there are a couple of things that we can change to make it a more flexible and semantically accurate component. We can take a look together if there's time after the meeting tomorrow. Also it'll need to be in the main addon package.

tests/dummy/app/components/card-pay/suggested-next-steps-box/index.hbs
```diff
@@ -7,11 +7,11 @@
     </span>
   </div>
   <div class="card-pay-suggested-next-steps-box__content">
-  <div class="card-pay-suggested-next-steps-box__cluster">
-    <div class="card-pay-suggested-next-steps-box__cluster-body">
+  <ul class="card-pay-suggested-next-steps-box__cluster">
```
Aierie: @burieberry realised just now that I made a mistake here - there is basically only one li and all the buttons are rendered in it. I will correct this after you've pushed changes, but tagging you here first just in case.

burieberry: @Aierie So I ended up combining our ideas in the `Boxel::NextStepsBox` component. See what you'll think of it. It turns out `if (has-block @block)` is fine to use with named blocks. I only edited the `ul` part, using ideas from this [slide](https://speakerdeck.com/chancancode/virtual-emberconf-2021-platform-state-of-the-union?slide=33) from ember conf. Also added a `default` block which can render if any additional content is needed. I kept the cluster css. We can do it this way until `gap` for flexbox is more widely supported by browsers.

Aierie: I like this! Thanks for making these changes. Have removed the card-pay component since it doesn't provide much value over the Boxel one, and updated some description text.

PR #329, opened 2021-04-02
fix some typescript declaration error messages

Aierie:
Changes:
1. Only one * for wildcard module declarations
2. Move the global webkit audio declaration to a separate file to stop it from breaking other declarations with the `export`
3. Added a few declaration types for `svg` and `flac`
4. Remove `export default config` from the `environment.d.ts` which doesn't seem to be needed
5. Change `Buffer` -> `ArrayBuffer` to fix error message

I don't fully understand what is happening here, will do a bit more reading, but putting this here for visibility first.


lukemelia (approved):
Ahhhh, good job!

PR #339, opened 2021-04-08
Feature/custom radio

Aierie:
This is a custom radio component generated along the way while trying to create a Web3Modal-like UI component. This exists because: 

1. Custom radios might come up again 
2. Custom checkboxes might too, which seem like they will use similar code

And this might save some time on research + testing accessibility. 

The core idea behind this is that you can use `Boxel::CustomRadio::Item` inside `Boxel::CustomRadio::Container` and both these elements will behave like a generic container/div for layout purposes, while retaining accessibility that you get from using the correct html elements for a "pick one of many" scenario.

There might be a good, cross-browser way to do this with aria, but the current component simply puts radio inputs in a fieldset. Have tested with VoiceOver on Safari - for text content which contains block elements, it's probably better with `aria-labelledby` than `for` because with `aria-labelledby` VoiceOver actually reads all the content regardless of block/inline, while `for` makes VoiceOver read only the first block content and then say "and x more".

Am aiming for a base that is good enough to use for now, and that we can build on later. It is good to test with a screenreader when using this component to know how our content is being read, I think. Ideally we would be able to test on Windows with the popular screen readers there as well, but right now that's a bit hard.


Aierie:
Added a demo for the wallet connection card. This is a radio selection using the `CustomRadio` component. The commits that start with EXTRACT! will be removed from this PR.
image

addon/components/boxel/custom-radio/container/index.hbs
```diff
@@ -1,4 +1,7 @@
-<fieldset class="boxel-custom-radio__container" ...attributes>
+<fieldset class="boxel-custom-radio__fieldset">
```
Aierie: I can imagine some problems because of this special case fix - eg. trying to set attributes including a class on the containing fieldset but failing. This fix (wrapping a div) should not be here long term, but does not affect its immediate use case (the wallet selection). We should probably move towards an aria-based solution when possible.

Aierie (commented):
See https://www.w3.org/WAI/tutorials/forms/grouping/ for high-level context on current fieldset implementation + aria alternatives.

burieberry (approved):

PR #341, opened 2021-04-12
loading button

Aierie:
CS-560

Adds loading state to button. 

image

In the loading state, the button is not interactive. A loading spinner is added by the component, on the left side of the button - within the button's layout flow so that it can be centered.

### Some possible follow ups
1. Convert `@loading` and `@disabled` to a single argument `@state` - since you probably shouldn't have a loading AND disabled button? Right now it works though.
2. Change the button to accept a `@onClick` argument. This allows better interception of actions on the button when loading. 

### Why not just add the `disabled` attribute?
Screen reader jumps from the button to other things: https://gomakethings.com/dont-disable-buttons-while-submitting-forms-with-ajax/

### DIFF
--- a/addon/components/boxel/button/index.css
+++ b/addon/components/boxel/button/index.css
@@ -28,20 +28,41 @@
   letter-spacing: var(--boxel-button-letter-spacing, var(--boxel-lsp-lg));
 }
 
+.boxel-button__loading-indicator {
+  --icon-color: var(--boxel-button-text-color);
+  width: var(--boxel-button-loading-icon-size);
+  height: var(--boxel-button-loading-icon-size);
+  margin-right: var(--boxel-sp-xxxs);
+}
+
 /* select disabled buttons and links that don't have href */
 /* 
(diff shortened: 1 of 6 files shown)


lukemelia:
@Aierie Do we need to guard against non-pointer interactions with the button? i.e. tab focus plus enter?

Aierie:
> @Aierie Do we need to guard against non-pointer interactions with the button? i.e. tab focus plus enter?

Currently guarding against it using negative tabindex when it's loading

Aierie:
I tested with a screenreader. negative tabindex and pointer-events none does not guard against activation of a button via screenreader. Will have to change this - the easiest way looks to be disabling the button for now.

An `@onClick` argument could also do this, do you think it's good to move to this API? @burieberry @lukemelia

lukemelia:
> An @onClick argument could also do this, do you think it's good to move to this API? @burieberry @lukemelia

I think ultimately this gives us more control and is the right move.

Aierie:
Merging for now. Will handle screen reader disabling during loading later - ideally with `@onClick`.

PR #343, opened 2021-04-13
Feature/cta block

Aierie:
CS-569

image


Aierie:
States that this aims to cover:
image
image
image
image
image
image

Aierie:
This component feels like it's doing too much. Possibly could split sizing/padding + step into a separate component and focus solely on the contents.

addon/components/boxel/cta-block/index.ts
```diff
@@ -0,0 +1,157 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { equal, reads } from 'macro-decorators';
+
+enum CtaBlockState {
```
Aierie, line 5: States and sections are documented via comments here.

addon/components/boxel/cta-block/index.ts
```diff
@@ -0,0 +1,157 @@
...
+interface CtaBlockStateDescription {
+  // text on the main action button/the done status text
+  text: string;
+  // action that is called when the main action button is pressed
+  action?: () => void;
+  // text on the cancel button
+  cancelText?: string;
+  // action that is called when the cancel button is pressed
+  cancelAction?: () => void;
+}
```
Aierie: States and sections documented via comments here

Aierie:
Wondering if it would be good to have a call with design to talk about this component and the design constraints we should work with.

Aierie:
If implementation is ok, will add tests before merging. Tests will mainly check if for each state: 
- the correct ui elements and styling is shown for combinations of `stepNumber`, `canEdit`, `canCancel`, and the `yield`ed block
- The correct text appears on the main action, the cancel button, and the step number, and the correct callback is called

Aierie:
Will be modifying this based on conversation w @burieberry @lukemelia, aiming to move this towards a simpler named-blocks API that allows easier customization.

Some design constraints from brief chat with design team:
1. One action per CTA, for now.
2. We won't show a CTA in the memorialized state if there is none, so we don't need to make a disabled state for the memorialized version
3. Padding is done on a case-by-case basis right now.

Aierie:
Focus is being thrown off because we are replacing entire HTML elements rather than changing styles. Holding off on adding aria-live, because VoiceOver announces all of the Cta's content. "Loading Cancel actions only visible to you" might be more confusing than it is helpful. There should be a better way to do this, so I will add this and other accessibility issues to the follow-up Linear issue.

addon/components/boxel/cta-block/usage.hbs
```diff
@@ -0,0 +1,177 @@
...
+                <li>in-progress</li>
+              </ul>
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+  </:description>
+
+  <:example>
```
lukemelia, line 188: I think it would be helpful to show an example of low-profile usage -- i.e. just using the default named block.

addon/components/boxel/cta-block/index.hbs
```diff
@@ -0,0 +1,96 @@
...
+    </div>
+  {{/if}}
+  {{!-- 
+    We cannot pass splattribute classes to components in the component helper: 
+    https://github.com/emberjs/ember.js/issues/18402 
+  --}}
+  {{#if (and (has-block "default") this.isDefault)}}
+    {{yield 
+      (hash 
+        MainActionButton=(component
```
lukemelia: Could we shorten the yielded component names to "ActionButton", "CancelButton" and "Info"? It seems unnecessarily verbose currently.

lukemelia (approved):
Excellent improvement, overall. Left an API consideration and a doc suggestion.

addon/components/boxel/cta-block/index.hbs
Aierie: I think shortening the button names is an improvement, but would prefer to keep `Area` for the sake of showing that it's just a container that you can put anything in. If we change the button names, I will also change `MainActionArea` to `ActionStatusArea`.

lukemelia: Sounds reasonable

burieberry (approved):

PR #346, opened 2021-04-14
Introduce stylelint

lukemelia:


tests/dummy/app/css/components/isolated-collection.css
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable */
```
Aierie, line 1: Here

tests/dummy/app/css/components/version-history.css
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable */
```
Aierie, line 1: Here

Aierie (approved):
Thanks for working on this! This looks good, doesn't look like stylelint aggressively introduces new changes/practices. Having blocks of declarations move around was a little jarring in some places but I think it's a good thing so we can read css from top to bottom with the expectation that the bottom will take priority (https://stylelint.io/user-guide/rules/no-descending-specificity).

Why did we have stylelint disabled for these 2 files?

lukemelia:
>Why did we have stylelint disabled for these 2 files?

Ah, yes. Those files had so many dozens of lint complaints that I didn't have time to start untangling them!

Aierie:
> > Why did we have stylelint disabled for these 2 files?
> 
> Ah, yes. Those files had so many dozens of lint complaints that I didn't have time to start untangling them!

I see! Seems okay to leave them as is considering they're not Boxel components.

PR #349, opened 2021-04-18
Enhancement/date helpers

Aierie:
CS-507

Replaces moment-based helpers with dayjs-based helpers, reintroduces ember-beta to CI/CD. This is intended as a stop-gap so that we can still run CI/CD with ember-release when ember does a new release. We can introduce a proper date/time addon at a later date.

Implementation of helpers is largely copied from `ember-moment` so that we use their API, though this PR does not include `ember-moment`'s built-in timezone and locale support.


Aierie:
Found a library with a goal to be a drop-in replacement for ember-moment. It doesn't have proper tests yet and doesn't handle positional parameters for `moment-format` the same way, but I think it's nicely structured and is a good base to start with. It comes with timezone and locale support + other plugins included.

https://github.com/sinankeskin/ember-dayjs

lukemelia (approved):
Looks good. I noticed one additional dependency that we could drop, which I've pushed to the branch.

PR #356, opened 2021-04-27
Enhancement/thread message layout

Aierie:
CS-638

Restructures layout of thread message component to follow design specs for workflows in Card Pay.

Adds two new arguments to thread messages to accommodate for cards + differences in space allocated between memorialized and default state cards:
- `@containsCard` - whether the thread message contains a card or not.
- `@fullWidth` - whether to allocate the full width to the card. Only works if `@containsCard` is true. This is for default state cards/cards that require action.

image

### DIFF
--- a/addon/components/boxel/thread-message/index.css
+++ b/addon/components/boxel/thread-message/index.css
@@ -31,16 +31,6 @@
   border-radius: initial;
 }
 
-.boxel-thread-message__initial {
-  width: var(--boxel-thread-message-avatar-size);
-  height: var(--boxel-thread-message-avatar-size);
-  border: var(--boxel-border);
-  border-radius: 100px;
-  display: flex;
-  align-items: center;
-  justify-content: center;
-}
-
 .boxel-thread-message__info {
   display: flex;
   align-items: center;
--- a/addon/components/boxel/thread-message/index.css
+++ b/addon/components/boxel/thread-message/index.css
@@ -1,9 +1,8 @@
 .boxel-thread-message {
-  --boxel-thread-message-avatar-size: calc(2 * var(--boxel-sp));
+  --boxel-thread-message-avatar-size: 2.5rem; /* 40px. Avatar should not be set to be larger than 60px or smaller than 20px. */
+  --boxel-thread-message-meta-height: 1.25rem; /* 20px */
   --boxel-thread-message-gap: var(--boxel-sp);
-  --boxel-thread-message-meta-height: var(--boxel-sp); /* set the height to the text height so that we can do layout based on this */
-
-  min-height: var(--boxel-thread-message-avatar-size); /* if meta is showing, we need to account for short heights which might (unlikely) lead to overlap with other text messages */
+  --boxel-thread-message-margin-left: calc(var(--boxel-thread-message-avatar-size) + var(--boxel-thread-message-gap));
 }
 
 .boxel-thread-message--hide-meta {
@@ -12,18 +11,14 @@
 
 .boxel-thread-message__meta {
   display: grid;
-  grid-template-columns: var(--boxel-thread-message-avatar-size) 1fr var(--boxel-thread-message-avatar-size);
+  grid-template-columns: var(--boxel-thread-message-avatar-size) 1fr;
   grid-template-rows: var(--boxel-thread-message-meta-height);
   align-items: start;
   gap: var(--boxel-thread-message-gap);
 }
 
-.boxel-thread-message--card .boxel-thread-message__meta {
+.boxel-thread-message--full-width .boxel-thread-message__meta {
   align-items: center;
-}
-
-.boxel-thread-message__avatar {
-  width: var(--boxel-thread-message-avatar-size);
 }
 
 .boxel-thread-message__avatar-img {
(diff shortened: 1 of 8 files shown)


lukemelia:
These new properties start to make the API feel brittle and precious. As a consumer of the API, I have to somehow remember to set @containsCard to true when I embed a card and then to somehow figure out how to wire @fullWidth together with the state of that card.

It makes me think it would be better to have `<Boxel::ThreadMessage>`, accepting a `@message` arg and `<Boxel::ThreadCard>`, accepting an `@component` arg.

Aierie:
> These new properties start to make the API feel brittle and precious. As a consumer of the API, I have to somehow remember to set @containsCard to true when I embed a card and then to somehow figure out how to wire @fullWidth together with the state of that card.
> 
> It makes me think it would be better to have `<Boxel::ThreadMessage>`, accepting a `@message` arg and `<Boxel::ThreadCard>`, accepting an `@component` arg.

Given the way we are using it with Card Pay, I think `Boxel::ThreadCard` sounds reasonable.

burieberry:
I think we should go back to an older version of the ThreadMessage component. Now I see why I had done it the way I did back then:

```
<div>
   <div>
     Avatar img
     Name and timestamp
   </div>
   
   {{yield}}
</div>
```
The yield part we can style in Card Pay. Let's not lock ourselves into messages being a certain size and cards being a certain size. On mobile view, the message is also full-width (not just the card).

burieberry:
Another option is to keep things as is but only add `@fullWidth` arg in Michael's PR. If it's full-width, then the block that contains the message or card gets `grid-column: 1 / -1;` property. Let's not do the message margins in this component. This component should only care about its own contents, the containing container cares about the relationship between the message components. So even `.thead-message + .thread-message { margin-top: x; ` should be removed from here.

Aierie:
I feel like this runs into problems at the seams when trying to formalize design specs across components. Dimensions ideally should be controlled by the card. But controlling width means you have to control layout, via margins or grid. This ideally should be controlled by the container - but given the conditional layout rules we have, this means the container (`Boxel::Thread`) needs to know some css classes of its child (`Boxel::ThreadMessage`).

burieberry:
I'm looking at this PR more carefully now that I'm more available, and it looks pretty similar to my suggestion actually. I think we should combine the @containsCard and @fullWidth to be one argument instead of two.

burieberry:
> I'm looking at this PR more carefully now that I'm more available, and it looks pretty similar to my suggestion actually. I think we should combine the @containsCard and @fullWidth to be one argument instead of two.

So basically you either get the message contents on a new line with full width allowance:
full-width

Or you get this:
not-full-width

The card already has the rest built-in.

Aierie:
> I'm looking at this PR more carefully now that I'm more available, and it looks pretty similar to my suggestion actually. I think we should combine the @containsCard and @fullWidth to be one argument instead of two.
image

`@containsCard` means that we need a `margin-top: 20px` for the contents of the message if there is meta information displayed. `@fullWidth` means that we allocate the full width to the message. If we do combine them, I guess we can control the width of the card manually within the message? But then that will require manually coordinating the contents with the dimensions of the avatar + message text.

Aierie:
> Let's not do the message margins in this component. This component should only care about its own contents, the containing container cares about the relationship between the message components. So even `.thead-message + .thread-message { margin-top: x; ` should be removed from here.

re: responsibilities for laying out items - yes, I think containers generally should do this work. I would like to nest `Boxel::ThreadMessage` within `Boxel::Thread` and not show `Boxel::ThreadMessage` as a top level component if we do this.

burieberry:
> > I'm looking at this PR more carefully now that I'm more available, and it looks pretty similar to my suggestion actually. I think we should combine the @containsCard and @fullWidth to be one argument instead of two.
> > image
> 
> `@containsCard` means that we need a `margin-top: 20px` for the contents of the message if there is meta information displayed. `@fullWidth` means that we allocate the full width to the message. If we do combine them, I guess we can control the width of the card manually within the message? But then that will require manually coordinating the contents with the dimensions of the avatar + message text.

Correct. Except instead of the last sentence, remove `grid-template-rows: var(--boxel-thread-message-meta-height);` restriction from the grid. And use the smallest margin above the thread content so it's closer to the avatar.

Aierie:
> > > I'm looking at this PR more carefully now that I'm more available, and it looks pretty similar to my suggestion actually. I think we should combine the @containsCard and @fullWidth to be one argument instead of two.
> > > image
> > 
> > 
> > `@containsCard` means that we need a `margin-top: 20px` for the contents of the message if there is meta information displayed. `@fullWidth` means that we allocate the full width to the message. If we do combine them, I guess we can control the width of the card manually within the message? But then that will require manually coordinating the contents with the dimensions of the avatar + message text.
> 
> Correct. Except instead of the last sentence, remove `grid-template-rows: var(--boxel-thread-message-meta-height);` restriction from the grid. And use the smallest margin above the thread content so it's closer to the avatar.

I'm not sure what this means - the `grid-template-rows: var(--boxel-thread-message-meta-height)` restricts the height to handle the text case, if we want to add avatar and meta within the same containing element. The margins are currently done with measurements requested by design, so moving closer might not be correct for this.

I was thinking more about this case:
image
where we have a card, so we add `@containsCard`, but then that makes the contents full width too. The card isn't full width, so we have to constrain its width, then we have to coordinate the width and layout within the message so that the card lines up with the meta and/or text above it.

Aierie:
I think manual adjustment / placement of contained cards is ok for now, but would prefer having a way to make it explicit that cards should be placed in certain ways - adding a grid is one way, but grid is not great if multiple elements are added to one `Boxel::ThreadMessage`, and requires consumers making certain assumptions in code about the implementation of the grid if we want to use the grid to control sizes and placement. We can solve this problem later though.

PR #360, opened 2021-05-04
add on missing splattributes and @class

Aierie:
Fixes some styling for https://github.com/cardstack/cardstack/pull/1584


lukemelia:
The historical reason for the `@class` usage was for when we were using a component with the `component` helper. Just want to double-check whether that applies in these cases at all?

Aierie:
> The historical reason for the `@class` usage was for when we were using a component with the `component` helper. Just want to double-check whether that applies in these cases at all?

I can't find any at the moment, but wondering if buttons will need this, and if this should apply to `Boxel::Button`.

lukemelia (approved):

PR #364, opened 2021-05-10
Add autoscroll modifier and use for Boxel::Thread content

lukemelia:


addon/components/boxel/thread/index.hbs
```diff
@@ -4,7 +4,7 @@
   </div>
 
   <div class="boxel-thread__content-wrapper">
-    <div class="boxel-thread__scroll-wrapper" tabindex="0">
+    <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll}}>
```
Aierie, line 7: Should we allow the `lockThreshold` to be set as an argument?

Aierie (approved):
I think this works, with a few quirks. Watching the subtree means stateful cards with conditionally rendered items + sidebar changes can trigger autoscrolling, though with the right threshold this shouldn't be too noticeable. 

It might be possible to change the layout implementation so that the scrolling container only contains thread messages, which would solve both these problems, since we can remove the subtree watching.

Aierie:
Wondering about how this ties in with the notice about new messages. I wonder if scrolling to the bottom should be handled there as well (when a new message comes in, either add a notice at the top or scroll to the bottom - handled within the same function).

image

lukemelia:
> Wondering about how this ties in with the notice about new messages. I wonder if scrolling to the bottom should be handled there as well (when a new message comes in, either add a notice at the top or scroll to the bottom - handled within the same function).

I thought we'd cross that bridge when we come to it. AFAIK, we don't yet have design for this.

addon/components/boxel/thread/index.hbs
lukemelia: It is supported. In this case, we are using the default value of 10px.

PR #365, opened 2021-05-17
NextStepsBox - Remove `item` named block, apply cluster layout

Aierie:
CS-788

Removes the named block <:item> from `<NextStepsBox>` to prevent usage of a boilerplate-heavy pattern that can be replaced with `display: flex; gap: your-gap-here`, since flex gap has support in the latest versions of Chrome, Firefox, and Safari.


lukemelia (approved):
Is the idea that we'll use a button-container div and the same CSS everytime we use `<Boxel::NextStepsBox>`?

Aierie:
> Is the idea that we'll use a button-container div and the same CSS everytime we use `<Boxel::NextStepsBox>`?

With this implementation and current design, yes. We could also apply that layout here by default, since right now all of the `NextStepsBox`es use that layout with set dimensions, but I held off on that and left the contents as a simple div in case design spec changes.

lukemelia:
> I held off on that and left the contents as a simple div in case design spec changes

Let's keep the common case simple.

burieberry (approved):
Works for me. FYI the second usage example in the staging link is still showing up, but I see that it's removed in your code. Must be some deployment delay.

Aierie:
> Works for me. FYI the second usage example in the staging link is still showing up, but I see that it's removed in your code. Must be some deployment delay.

No, that was my mistake. Thanks for spotting that @burieberry! Removed more unused things from the file.

lukemelia (approved):

PR #366, opened 2021-05-17
Feature/loading indicator

Aierie:
CS-711

Brings the animated loading indicator from CardPay into Boxel as a component, and uses it in other components that were previously using the svg directly. The component includes sizing, animation, and an argument to specify the color.

### DIFF
--- a/addon/components/boxel/loading-indicator/index.css
+++ b/addon/components/boxel/loading-indicator/index.css
@@ -1,0 +1,18 @@
+/* zero specificity default sizing */
+:where(.boxel-loading-indicator) {
+  width: 1.25rem;
+  height: 1.25rem;
+}
+
+/* 
+  Only animate if the user has not said that they want reduced motion
+*/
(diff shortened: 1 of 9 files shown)


addon/components/boxel/loading-indicator/index.css
```diff
@@ -0,0 +1,18 @@
+/* zero specificity default sizing */
+:where(.boxel-loading-indicator) {
+  width: 1.25rem;
+  height: 1.25rem;
+}
```
lukemelia, line 5: Had to look this up. Interesting!

lukemelia (approved):
LGTM

addon/components/boxel/loading-indicator/index.css
Aierie: Yeah, useful so that we don't have to double up the selectors as consumers! Luckily the specificity issue came up immediately when working on components that use this and this seemed like a reasonable fix.

PR #369, opened 2021-05-27
Input field optional and error states styling

burieberry:
CS-497

Adds `@helperText`, `@errorMessage`, `@optional` args to Boxel::Input component and their styling

required-field

### DIFF
--- a/addon/components/boxel/input/index.hbs
+++ b/addon/components/boxel/input/index.hbs
@@ -11,7 +11,7 @@
       required={{@required}}
       disabled={{@disabled}}
       aria-describedby={{if @helperText (concat "helper-text-" helperId) false}}
-      aria-invalid={{if @invalid true false}}
+      aria-invalid={{if @invalid "true"}}
       aria-errormessage={{if shouldShowErrorMessage (concat "error-message-" helperId) false}}
       data-test-boxel-input
       data-test-boxel-input-id={{@id}}
--- a/addon/components/boxel/input/index.hbs
+++ b/addon/components/boxel/input/index.hbs
@@ -1,23 +1,27 @@
 {{#if (and (not @required) @optional)}}
   <div class="boxel-input__optional">Optional</div>
 {{/if}}
-<Input
-  class={{cn "boxel-input" boxel-input--invalid=@invalid}}
-  id={{@id}}
-  @value={{@value}}
-  {{on "input" (optional @onInput)}}
-  required={{@required}}
-  disabled={{@disabled}}
-  aria-describedby={{if @helperText (concat "helper-text-" @id) false}}
-  aria-invalid={{if @invalid true false}}
-  aria-errormessage={{if @errorMessage (concat "error-message-" @id) false}}
-  data-test-boxel-input
-  data-test-boxel-input-id={{@id}}
-  ...attributes
-/>
-{{#if @errorMessage}}
-  <div id={{concat "error-message-" @id}} class="boxel-input__error-message">{{@errorMessage}}</div>
-{{/if}}
-{{#if @helperText}}
-  <div id={{concat "helper-text-" @id}} class="boxel-input__helper-text">{{@helperText}}</div>
-{{/if}}
+{{#let (and @invalid @errorMessage) as |shouldShowErrorMessage|}}
+  {{#let (unique-id) as |helperId|}}
+    <Input
+      class={{cn "boxel-input" boxel-input--invalid=@invalid}}
+      id={{@id}}
+      @value={{@value}}
+      {{on "input" (optional @onInput)}}
+      required={{@required}}
+      disabled={{@disabled}}
+      aria-describedby={{if @helperText (concat "helper-text-" helperId) false}}
+      aria-invalid={{if @invalid true false}}
+      aria-errormessage={{if shouldShowErrorMessage (concat "error-message-" helperId) false}}
+      data-test-boxel-input
+      data-test-boxel-input-id={{@id}}
+      ...attributes
+    />
+    {{#if shouldShowErrorMessage}}
+      <div id={{concat "error-message-" helperId}} class="boxel-input__error-message" data-test-boxel-input-error-message>{{@errorMessage}}</div>
+    {{/if}}
+    {{#if @helperText}}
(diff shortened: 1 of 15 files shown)


addon/components/boxel/input/index.hbs
```diff
@@ -1,11 +1,23 @@
...
   data-test-boxel-input-id={{@id}}
   ...attributes
-/>
\ No newline at end of file
+/>
+{{#if @errorMessage}}
+  <div id="error-message" class="boxel-input__error-message">{{@errorMessage}}</div>
+{{/if}}
+{{#if @helperText}}
+  <div id="helper-text" class="boxel-input__helper-text">{{@helperText}}</div>
```
Aierie: Using multiple inputs means that we have multiple elements with the id 'helper-text' on the same page if they all have helper text. If we want to use aria-describedby, we need to enforce unique ids for each input's helper text. Same for elements with the id error-message. 

One way of doing this could be assigning a unique ID on a per-component basis, and appending that to 'helper-text' and 'error-message'?

addon/components/boxel/input/usage.hbs
```diff
@@ -4,9 +4,13 @@
     <Boxel::Input
       @id={{this.id}}
       @value={{this.value}}
-      @onInput={{@onInput}}
+      @onInput={{this.validate}}
       @disabled={{this.disabled}}
       @required={{this.required}}
+      @optional={{this.optional}}
+      @errorMessage={{this.errorMessage}}
```
Aierie: `@required` adds the `required` html attribute, but doesn't prevent users from submitting stuff unless we add an action similar to `this.validate` for every component that uses this input, or use the input within a form. It also doesn't add any error messages unless it's specified by the parent. I would expect a `@required` argument to do one of the above, but I don't think it does without some help at the moment.

Aierie: The main value the `@required` argument adds is preventing the `@optional` label from being shown. Is preventing conflicts between `@required` and `@optional` the job of the user and not this component? If so, would it be better to simply supply `required` via a html attribute?

About the label that currently reads "Optional": 
1. Can we use `not @required` to tell when to show the label? This might be possible if we always show the "Optional" label when a field is not required.
2. Is a string argument better for the label? Eg. if we require some flexibility, eg. "Only for merchants", or using the label when a field is required.

image

Aierie: Also think it's better to control whether the error message shows via a separate toggle, eg. `@invalid`. Otherwise, all components that validate inputs will have to replace error messages with falsey values until they actually want to show them.

Aierie (changes requested):

addon/components/boxel/input/usage.hbs
burieberry: I think it's fine for now to leave it up to the user whether they want to show the `Optional` label. There are cases where the field is optional but we don't want to display the optional label, a search field for example. Also I think it's fine to leave it up to the user what they want their error validation logic to be. It'll be different for different types of input field.

I started out with separate args for @invalid and @errorMessage, but then decided that to be redundant. The @errorMessage is falsey by default, and if an entry is deemed invalid, there should be a message explaining why.

I like the idea of making the optional field label a string field.

Aierie: Always showing a useful error message is nice! I'm for having `@invalid`  because it makes it easier to control when to show an error message, instead of having to assign falsey values to validation errors until they need to be shown, if we use getters to show them.

```
<Boxel::Input
  @value={{this.value1}}
  @errorMessage={{this.errorMessage1}}
/>
<Boxel::Input
  @value={{this.value2}}
  @errorMessage={{this.errorMessage2}}
/>
<button {{on "click" (fn (mut this.showErrors) true) }}>Show errors</button>

/* In js */
export default class Something extends Component {
  @tracked value1 = "";
  @tracked value2 = "";
  @tracked showErrors = false;

  get errorMessage1(){
    if (!this.showErrors) return ''
    if (this.value1.length < 20) return 'Answer is too short'
  }

  get errorMessage2(){
    if (!this.showErrors) return ''
    if (this.value2.length < 60) return 'Answer is too short'
  }
}
```

If we have `@invalid`, this becomes

```
<Boxel::Input
  @value={{this.value1}}
  @errorMessage={{this.errorMessage1}}
  @invalid={{this.showErrors}}
/>
<Boxel::Input
  @value={{this.value2}}
  @errorMessage={{this.errorMessage2}}
  @invalid={{this.showErrors}}
/>
<button {{on "click" (fn (mut this.showErrors) true) }}>Show errors</button>

/* In js */
export default class Something extends Component {
  @tracked value1 = "";
  @tracked value2 = "";
  @tracked showErrors = false;

  get errorMessage1(){
    if (this.value1.length < 20) return 'Answer is too short'
  }

  get errorMessage2(){
    if (this.value2.length < 60) return 'Answer is too short'
  }
}
```

I'm not sure how to force users to provide an error message while allowing the invalid state to be triggered though.

burieberry:
Decided to keep the `@optional` label as is, any explanation can be done in the `@helperText` field. Added `@id` arg to aria fields for `@helperText` and `@errorMessage`. Passing `@invalid` as a separate arg.

Aierie:
Rebased to fix failing CI.

Aierie:
Do you know what happens if this input is wrapped in a label, and the helper, error text, and the optional label is shown? Does it get read as part of the label? @burieberry

Aierie:
Just realised this input seems to assume use within Boxel::Field - it uses grid positions that help for styling within the Boxel::Field component. This works for immediate use, but introduces unexpected behaviour when used within a grid for other things, and makes it dependent on Boxel::Field styling. 

Since helper and error text being included in a wrapping label is also an issue, I'm thinking we could:
1. Limit the responsibility of the Boxel::Input component. It controls the state - hover, focus, invalid. Wrapping components (eg. Boxel::Field) are responsible for ensuring an appropriate label, and display and placement of the optional label, error messages and helper text, including corresponding aria. Things like aria and matching ids can be implemented by using Boxel::Input as a contextual component.
2. Or we could make this input component have a built-in label that displays itself like a Boxel::Field. 

Am leaning more towards option 1.

Aierie (approved):

PR #372, opened 2021-06-01
Enhancement/disabled action chins

Aierie:
Work to support https://github.com/cardstack/cardstack/pull/1659

- Removes the disabled named block and state from the action chin
  - For all uses in the monorepo so far, the disabled named block is the same as the default block besides the button being disabled
  - Only exception is the styling for disabled action chins with steps
- Adds the `@disabled` argument that causes all contextual buttons from this component to be disabled by default, and styling for disabled action chins with steps to be applied
- Fixes some contrast issues in disabled action chins

### DIFF
--- a/addon/components/boxel/action-chin/index.css
+++ b/addon/components/boxel/action-chin/index.css
@@ -46,12 +46,12 @@
   --boxel-action-chin-padding-vertical: var(--boxel-sp-xl);
 }
 
-.boxel-action-chin--memorialized.boxel-action-chin--has-step {
+.boxel-action-chin--memorialized.boxel-action-chin--has-step,
+.boxel-action-chin--disabled.boxel-action-chin--has-step {
   --boxel-action-chin-padding-vertical: var(--boxel-sp-xs);
 }
 
-.boxel-action-chin--disabled.boxel-action-chin--has-step {
-  --boxel-action-chin-padding-vertical: var(--boxel-sp-xs);
+.boxel-action-chin--default.boxel-action-chin--disabled.boxel-action-chin--has-step {
   --boxel-action-chin-background-color: var(--boxel-purple-700-opacity90);
 }
 
--- a/addon/components/boxel/action-chin/index.hbs
+++ b/addon/components/boxel/action-chin/index.hbs
@@ -2,6 +2,7 @@
   class={{cn
     "boxel-action-chin"
     (if @state (concat "boxel-action-chin--" @state))
+    (if @disabled "boxel-action-chin--disabled")
     (if @stepNumber (concat "boxel-action-chin--has-step"))
   }}
   ...attributes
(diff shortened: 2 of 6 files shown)


addon/components/boxel/action-chin/index.css
```diff
@@ -4,7 +4,7 @@
   --button-base-height: 2rem; /* The height of the boxel button, used for calculating the min-height to reduce layout shift */
   --boxel-action-chin-background-color: var(--boxel-purple-700);
   --boxel-action-chin-emphasis-text-color: var(--boxel-light);
-  --boxel-action-chin-text-color: var(--boxel-purple-300);
+  --boxel-action-chin-text-color: var(--boxel-light-500);
```
burieberry: I've actually checked the contrast here and it is sufficient, so no need to change

addon/components/boxel/action-chin/index.css
```diff
@@ -29,7 +29,7 @@
 .boxel-action-chin--memorialized {
   --boxel-action-chin-background-color: var(--boxel-purple-100);
   --boxel-action-chin-emphasis-text-color: var(--boxel-dark);
-  --boxel-action-chin-text-color: var(--boxel-purple-400);
+  --boxel-action-chin-text-color: var(--boxel-purple-700);
```
burieberry: This one also has sufficient contrast

contrast

addon/components/boxel/action-chin/index.css
```diff
@@ -46,13 +46,14 @@
...
+.boxel-action-chin--memorialized.boxel-action-chin--has-step,
+.boxel-action-chin--disabled.boxel-action-chin--has-step {
   --boxel-action-chin-padding-vertical: var(--boxel-sp-xs);
 }
 
-.boxel-action-chin--disabled.boxel-action-chin--has-step {
-  --boxel-action-chin-padding-vertical: var(--boxel-sp-xs);
+.boxel-action-chin--default.boxel-action-chin--disabled.boxel-action-chin--has-step {
   --boxel-action-chin-background-color: var(--boxel-purple-700-opacity90);
+  --boxel-action-chin-text-color: var(--boxel-light-200);
```
burieberry: This color change is not necessary

burieberry (commented):
I'm fine with removing the `disabled` state's block. Looks good. A made a comment in a couple of places where I don't think color change is necessary, since we're AA compliant. If it's because you're having trouble with the a11y test, it gets confused by the css-transitions I think.

Aierie:
> I'm fine with removing the `disabled` state's block. Looks good. A made a comment in a couple of places where I don't think color change is necessary, since we're AA compliant. If it's because you're having trouble with the a11y test, it gets confused by the css-transitions I think.

It is indeed the a11y test, which frustratingly fails in CI/CD while passing on local.

Aierie:
Removed a11yAudit for disabled action chins, since colors pass AA contrast and test results are inconsistent on different machines (CI/CD, and different Macbooks on the team).

burieberry (approved):

PR #375, opened 2021-06-03
Enhancement/native html input

Aierie:
- Replaced Boxel::Input internal input with native html input. 
- Added `@onInput` as a convenience argument that receives the input's value


addon/components/boxel/input/index.css
```diff
@@ -27,6 +27,11 @@
   box-shadow: 0 0 0 1px var(--boxel-error-100);
 }
 
+.boxel-input--invalid:focus {
+  outline: 1px solid transparent;  /* Make sure that we make the invalid state visible */
+  box-shadow: 0 0 0 1.5px var(--boxel-error-100);
+}
```
burieberry, line 33: Nice, this was starting to bother me

addon/components/boxel/input/index.ts
```diff
@@ -0,0 +1,12 @@
...
+
+interface BoxelInputComponentArgs {
+  onInput?(value: string): void;
+}
+
+export default class BoxelInputComponent extends Component<BoxelInputComponentArgs> {
+  @action handleInput(event: InputEvent): void {
+    this.args.onInput?.((event.target as HTMLInputElement).value);
+  }
+}
```
burieberry: This is not so difficult to accomplish via the `pick` helper or via destructuring. But if you want to keep it, we can keep. I don't see a downside of keeping it. Only makes this not a template-only component, which is fine.

burieberry (approved):

addon/components/boxel/input/index.ts
Aierie: Used the pick helper. Thanks!

PR #377, opened 2021-06-14
add an autoscroll argument to the thread component

Aierie:
To help with `prefers-reduced-motion: reduce`, allow consumers to turn off autoscrolling in the thread component. For use in https://github.com/cardstack/cardstack/pull/1697

### DIFF
--- a/addon/components/boxel/thread/index.hbs
+++ b/addon/components/boxel/thread/index.hbs
@@ -4,7 +4,7 @@
   </div>
 
   <div class="boxel-thread__content-wrapper">
-    <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll}}>
+    <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll disable=(not @autoscroll)}}>
       <div class="boxel-thread__sticky-container">
         {{yield to="taskbar"}}
       </div>
--- a/addon/components/boxel/thread/index.hbs
+++ b/addon/components/boxel/thread/index.hbs
@@ -4,7 +4,7 @@
   </div>
 
   <div class="boxel-thread__content-wrapper">
-    <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll disable=(not @autoscroll)}}>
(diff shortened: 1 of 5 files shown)


addon/components/boxel/thread/index.hbs
```diff
@@ -4,7 +4,7 @@
   </div>
 
   <div class="boxel-thread__content-wrapper">
-    <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll}}>
+    <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll disable=(not @autoscroll)}}>
```
lukemelia: This code makes me wish for a way to express this more elegantly (and by that I mean, without a double negative). `enabled=@autoscroll` might be better? The modifier would need to explicitly distinguish between not passing `enabled` vs specifying false, so I understand why you took this path.

lukemelia (approved):

addon/components/boxel/thread/index.hbs
Aierie: Will change this to enabled. Forgot about `Object.prototype.hasOwnProperty` for a bit.

lukemelia: The nullish coalescing operator (`??`) can be useful for this kind of thing too.

Aierie: Yep! Though it does mean passing `undefined` and `null` leads to autoscrolling. Was concerned it might lead to unexpected behaviour so opted for `hasOwnProperty` instead, which allows `undefined` and `null` to be understood as falsey values.

Aierie: `hasOwnProperty` makes `ember-source@3.27` unhappy in tests, using `Reflect.ownKeys` to figure out if `enabled` is provided instead.

Source that is throwing:
https://github.com/glimmerjs/glimmer-vm/blob/4f1bef0d9a8a3c3ebd934c5b6e09de4c5f6e4468/packages/%40glimmer/manager/lib/util/args-proxy.ts#L61

